PR #2, opened 2017-03-14
vendor: Add Vault API deps

rossf7:
- Add vault/api deps for Vault 0.6.5


w0mbat:
Do we need the vendor folder in there?

JosephSalisbury:
How else would we add dependencies?

rossf7:
Afraid so even with golang/dep when its ready!  :)

w0mbat:
@JosephSalisbury Still always wondering, why glide/golang has to package all deps into the repo - coming from the Java ecosystem makes this feel weird but its offtopic ;)

rossf7:
@w0mbat I used to do a lot of Ruby so vendor/ is also a bit strange for me ;)

The convention is to commit vendor/ and I'm trying to follow the structure of aws-operator and kvm-operator where possible.

PR #4, opened 2017-03-14
Add Kubernetes API client and daemon flags

rossf7:

* Use latest microkit with viper for flags.
* Add k8s client using k8s.io/client-go and only allow client cert auth.
* Add more missing deps.


client/k8s/error.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	"github.com/juju/errgo"
+)
+
+var (
+	maskAny = errgo.MaskFunc(errgo.Any)
+)
+
+func maskAnyf(err error, f string, v ...interface{}) error {
```
xh3b4sd: This is not necessary. We can simply use `microerror.MaskAny` and `microerror.MaskAnyf`. Here the import is then `microerror "github.com/giantswarm/microkit/error"`. Kubernetesd for instance does it like this.

client/k8s/k8s.go
```diff
@@ -0,0 +1,81 @@
...
+			return nil, err
+		}
+	} else {
+		rawClientConfig = newRawClientConfig(config)
+	}
+
+	return rawClientConfig, nil
+}
+
+func NewClient(config Config) (kubernetes.Interface, error) {
```
xh3b4sd: When this k8s client magic is state of the art and we want to use this in different projects it would be cool to prevent the copy/paste and add some package to microkit.

main.go
```diff
@@ -6,18 +6,37 @@ import (
...
+	f           *flag.Flag = flag.New()
+	gitCommit   string     = "n/a"
+	name        string     = "cert-operator"
+	source      string     = "https://github.com/giantswarm/cert-operator"
 )
 
+// Flags is the global flag structure used to apply certain configuration to it.
+// This is used to bundle configuration for the command, server and service
+// initialisation.
+var Flags = struct {
```
xh3b4sd: This is not necessary anymore. Above you already create the new flag structure which is automatically set up. See line 19. 

```
f           *flag.Flag = flag.New()
```

main.go
```diff
@@ -34,13 +53,26 @@ func main() {
...
 			serviceConfig.Logger = newLogger
+			serviceConfig.Viper = v
+
+			k8sTlsClientConfig := k8sclient.TLSClientConfig{
+				CertFile: Flags.Kubernetes.TLS.CrtFile,
+				KeyFile:  Flags.Kubernetes.TLS.KeyFile,
+				CAFile:   Flags.Kubernetes.TLS.CAFile,
+			}
+			serviceConfig.K8sConfig = k8sclient.Config{
+				Address:         Flags.Kubernetes.Address,
```
xh3b4sd: The flags are not used like this with the viper configuration. Check Kubernetesd. It should look like this. 

```
config.Viper.GetString(config.Flag.Service.Kubernetes.Address)
```

Lets talk if this is kind of unclear.

client/k8s/k8s.go
rossf7: Cool totally agree. There was more copy/paste than I'd like.

client/k8s/error.go
rossf7: I think I'm OK. I'll look again at kubernetesd and fix but I'll shout if I get stuck.

rossf7: Removed

rossf7:
@xh3b4sd The flag handling should be fixed now. I also added the validation checks for the k8s api address that are in kubernetesd. Can you give it another review please.

PR #8, opened 2017-03-17
Add issue certificate to the create service

rossf7:

This PR allows us to issue client certs for guest clusters when a certificate TPO is created. Uses the [certctl](https://github.com/giantswarm/certctl/) library to issue the certs using the Vault PKI backend.

Currently just issues a test cert when the create service boots. This is because the certificate TPR is not yet integrated. Will also be extended to automatically create a PKI backend for the cluster if it does not exist yet.


service/create/cert_signer.go
```diff
@@ -0,0 +1,73 @@
...
+type issueResponse struct {
+	Certificate  string
+	PrivateKey   string
+	IssuingCA    string
+	SerialNumber string
+}
+
+// Issue generates a certificate using the PKI backend signed by the certificate
+// authority associated with the configured cluster ID.
+func (s *Service) Issue(config CertificateSpec) (issueResponse, error) {
```
xh3b4sd: This looks like the code in the certctl libs. I thought we can simply reuse this. There is no reason to copy/paste all of the code.

rossf7: @xh3b4sd yes it is the certctl code. I see now that it makes sense to reuse the library rather than copying across. I'll look into that now and redo the PR.

rossf7:
@xh3b4sd I've put more details in the PR and fixed the code to use certctl. Can you take another look to check if the direction is OK?

service/create/error.go
```diff
@@ -10,3 +10,10 @@ var invalidConfigError = errgo.New("invalid config")
 func IsInvalidConfig(err error) bool {
 	return errgo.Cause(err) == invalidConfigError
 }
+
+var keyPairNotFoundError = errgo.New("key pair not found")
+
+// IsKeyPairNotFound asserts keyPairNotFoundError.
+func IsKeyPairNotFound(err error) bool {
+	return errgo.Cause(err) == keyPairNotFoundError
+}
```
xh3b4sd: This is no longer necessary I guess. At least it is not used in the code anymore.

service/create/service.go
```diff
@@ -50,6 +61,10 @@ func New(config Config) (*Service, error) {
 		return nil, microerror.MaskAnyf(invalidConfigError, "logger must not be empty")
 	}
```
xh3b4sd: Cosmetic nitpicking. The empty line here is not necessary. We can simply have the conditionals all together unless there is a deeper reason behind grouping them.

service/create/error.go
rossf7: Good point! I've removed this and fixed the whitespace issue :)

PR #9, opened 2017-03-17
Setup PKI backend in Vault for the cluster automatically

rossf7:

This PR automates the manual setup of PKI backends in Vault for each cluster. Currently this is done via the [certctl](https://github.com/giantswarm/certctl/) CLI tool. First a check is made to see if there is a valid PKI backend and if not then a backend is created.

Currently the operator just checks a test cluster when it boots. This is because the certificate TPR is not yet integrated. Will be extended so the setup is triggered automatically when a new cert TPO is detected by the watch.


service/create/pki.go
```diff
@@ -0,0 +1,119 @@
...
+
+	microerror "github.com/giantswarm/microkit/error"
+
+	"github.com/giantswarm/certctl/service/pki"
+	"github.com/giantswarm/certctl/service/token"
+)
+
+const (
+	// TODO Choose a sensible value and move to ConfigMap for V1
+	CertificateAuthorityTTL = "1440h"
```
rossf7: The CA TTL needs to be longer than the Cert TTL. For v1 is it OK to just use a config map? When we handle checking for certs expiring I think we'll also need to handle the CA expiring.

service/create/pki.go
```diff
@@ -0,0 +1,119 @@
...
+	service, err := s.getTokenService()
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	// Create PKI policy TODO Use latest certctl with CreateConfig object.
+	return service.CreatePolicy(cert.ClusterID)
+}
+
+// Get the common name for the Cluster CA by removing the prefix
```
rossf7: For the cluster CA I'm just removing the prefix from the cert name e.g.
* Cert: api.test-cluster.g8s.eu-west-1.aws.test.private.giantswarm.io
* CA: test-cluster.g8s.eu-west-1.aws.test.private.giantswarm.io

service/create/service.go
```diff
@@ -94,17 +94,42 @@ func (s *Service) Boot() {
...
+			ClusterID:  "g8s-cert-test-eu-west-1",
+			CommonName: "api.cluster-test.g8s.eu-west-1.aws.test.private.giantswarm.io",
+			IPSANs:     []string{"10.0.0.4", "10.0.0.5"},
+			AltNames: []string{
+				"kubernetes",
+				"kubernetes.default",
+				"kubernetes.default.svc",
+				"kubernetes.default.svc.cluster.local",
+			},
+			AllowBareDomains: true,
```
rossf7: These alt names are from kubernetesd. Since they include kubernetes allow bare domains will need to be true.

service/create/service.go
```diff
@@ -94,17 +94,42 @@ func (s *Service) Boot() {
...
 			TTL:              "720h",
 		}
 
+		s.Config.Logger.Log("info", "Checking for PKI backend")
+		if !s.checkPKIBackend(cert.ClusterID) {
+			err := s.createPKIBackend(cert)
```
rossf7: At the moment the PKI setup just happens once the first time a new cluster is detected. But I think there's a problem with alt names. If a 2nd cert has different alt names these won't be set up and issuing the cert will fail.

In certctl setup is additive so we could just call createPKIBackend every time? Or maybe we need to check the configured alt names?

service/create/pki.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func (s *Service) createPKIPolicy(cert CertificateSpec) error {
+	service, err := s.getTokenService()
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	// Create PKI policy TODO Use latest certctl with CreateConfig object.
+	return service.CreatePolicy(cert.ClusterID)
```
rossf7: I need to update to the latest certctl which has the number of tokens to issue. But for V1 I think this should be 0. In a later version I think we'll need to generate a token and store as a g8s secret.

rossf7:
@xh3b4sd @w0mbat could I get some feedback on the PKI setup code. I've added comments for the things I'm unclear on.

service/create/pki.go
xh3b4sd: IMO the CA should never expire on its own. We should only actively renew it in case of security breaches. This should be super rare. What we need to renew frequently are the self signed certificates for a cluster, but this is a different story we should not care about that much right now. Anyway we should have decent default values for nearly all of the configuration the microservice cares about. So this should then be in a configmap like we have it in e.g. kubernetesd and other microservices. So lets not put this in constants here. WDYT?

service/create/pki.go
xh3b4sd: I think we should have a common base domain and several formats to derive the actual common names for the certificates from. In the cluster-service we already do something like this. See https://github.com/giantswarm/cluster-service/blob/master/kubernetes/configmap.yaml#L31. Here it is only that the base domain is copy/paste and not use in combination. This should be reworked when we get some time for boyscouting. I would say lets try to do that right here.

service/create/service.go
xh3b4sd: AFAIK the alternative names should nearly never change. Do you have an example in mind where it might become a problem? I cannot think about any from the top of my head.

service/create/pki.go
xh3b4sd: Uncertain right now. Lets care about this later when this works for you.

service/create/pki.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (s *Service) checkPKIBackend(clusterID string) bool {
+	service, err := s.getPKIService()
+	if err != nil {
+		return false
+	}
+
+	// Check PKI config.
+	mounted, err := service.IsMounted(clusterID)
+	if !mounted || err != nil {
```
xh3b4sd: Lets please have always proper error handling. When shit hits the fan we are really screwed by this code because we do not see anything.

service/create/pki.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/giantswarm/certctl/service/pki"
+	"github.com/giantswarm/certctl/service/token"
+)
+
+const (
+	// TODO Choose a sensible value and move to ConfigMap for V1
+	CertificateAuthorityTTL = "1440h"
+)
+
+func (s *Service) checkPKIBackend(clusterID string) bool {
```
xh3b4sd: Nitpicking. What does it check? What about `verifyPKISetup`?

service/create/pki.go
rossf7: Yes it makes sense that the CA shouldn't expire unless there is a breach. I'll move this to a config map.

service/create/pki.go
xh3b4sd: I also think that it would make sense to put this logic into the certctl service and use it here. This is a common functionality we otherwise copy/paste in case it is required in other places.

service/create/pki_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetClusterCA(t *testing.T) {
+	svc, _ := New(DefaultConfig())
+
+	tests := []struct {
+		name           string
+		certCommonName string
+		expectedCA     string
```
xh3b4sd: This is the common name of the CA, right? Would be cool if the variable name reflects that somehow.

service/create/pki.go
rossf7: Thanks, that's much nicer. I'll do the same thing here. Yes it would be nice to move to microkit or operator-kit when we get time for boyscouting.

service/create/pki_test.go
```diff
@@ -0,0 +1,85 @@
...
+			altNames:            []string{},
+			expectedDomainCount: 1,
+		},
+	}
+
+	for _, tc := range tests {
+		cert := CertificateSpec{
+			CommonName: tc.certCommonName,
+			AltNames:   tc.altNames,
+		}
```
xh3b4sd: FYI I tend to try to keep the interface of table driven tests clean. So the single test case defines the complete input for the function being tested. It makes reading the test and testing the actual function a bit more easy IMO.

service/create/service.go
```diff
@@ -94,17 +94,42 @@ func (s *Service) Boot() {
...
+			if err == nil {
+				s.Config.Logger.Log("info", fmt.Sprintf("Created backend for %s", cert.ClusterID))
+			} else {
+				s.Config.Logger.Log("info", fmt.Sprintf("Failed to create PKI backend: %v", err))
```
xh3b4sd: Lets please use the format placeholder `#%v` when logging errors like this. Then we see the details of the error's stack. IIRC the KInvolks had already issues with this.

service/create/service.go
rossf7: No I don't have an example and it's good to know that the alternative names shouldn't change often. In that case I'll leave it as is for now rather than hold things up.

service/create/service.go
xh3b4sd: Further, if there is an error now, the code flow goes on. We should rather stop here and retry after some time. There are also libs for exponential backoffs, but this can gladly be done later in a separate PR.

service/create/pki.go
rossf7: Cool I agree lets deal with generating tokens later.

rossf7:
@xh3b4sd thanks for reviewing. I'll update the PR to fix the rest of these issues.

I'll also see if I can move more of the logic to certctl. I need to make changes there anyway to make the repo public. Also currently I can only get it working with 0.6.0 of Vault which I think need to be fixed.

rossf7:
@xh3b4sd I think the updated PR addresses all the issues except the retry logic and the k8s configmap resource. I'd like to do these as separate PRs. Shall I raise issues for them?

xh3b4sd:
Lets definitely track issues on our boards. I will check the PR now.

main.go
```diff
@@ -106,6 +106,8 @@ func main() {
 	daemonCommand.PersistentFlags().String(f.Kubernetes.TLS.KeyFile, "", "Key file path to use to authenticate with Kubernetes.")
 	daemonCommand.PersistentFlags().String(f.Vault.Address, "", "Address used to connect to Vault.")
 	daemonCommand.PersistentFlags().String(f.Vault.Token, "", "Token used to authenticate against Vault.")
+	daemonCommand.PersistentFlags().String(f.Vault.PKI.CATTL, "", "TTL used to generate a new Cluster CA.")
+	daemonCommand.PersistentFlags().String(f.Vault.PKI.CommonNameFormat, "", "Common name used to generate a new Cluster CA.")
```
xh3b4sd: Super picky and not really important in this specific case but I want to share my thoughts on this topic now because it will become quite important soon. When we deal with configuration I would like to think twice on the structure. I made the experience over the last years that structuring configuration is super important to understand and maintain it. So I think it should be `f.Vault.PKI.CA.TTL` and `f.Vault.PKI.CommonName.Format`. Just as a side note, this goes towards the story of configuration of G8S installations in general and how we deal with this. See also https://github.com/giantswarm/giantswarm/issues/1030.

service/create/pki.go
```diff
@@ -0,0 +1,98 @@
...
+	"fmt"
+	"strings"
+
+	microerror "github.com/giantswarm/microkit/error"
+
+	"github.com/giantswarm/certctl/service/pki"
+	"github.com/giantswarm/certctl/service/token"
+)
+
+// Creates a PKI backend if one does not exist for the cluster.
```
xh3b4sd: Please see https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences. The comment should start with `setupPKIBackend ...`.

service/create/pki.go
```diff
@@ -0,0 +1,98 @@
...
+		config := pki.CreateConfig{
+			ClusterID:        cert.ClusterID,
+			CommonName:       caCommonName,
+			AllowedDomains:   getAllowedDomainsForCA(caCommonName, cert),
+			AllowBareDomains: cert.AllowBareDomains,
+			TTL:              s.Config.Viper.GetString(s.Config.Flag.Vault.PKI.CATTL),
+		}
+
+		s.Config.Logger.Log("debug", fmt.Sprintf("Creating PKI backend for cluster %s", cert.ClusterID))
+		return service.Create(config)
```
xh3b4sd: I find it weird that a private method calls a public one. Like, it should be the other way around. When I talked with Timo it turned out we want to have a CA and a cert operator at the end. So we should already start having different service packages in here so we can more easily factor them out later. Getting the design more right now is quite important I guess. So I suggest we have a `service/ca` service package which provides a `Create` method and all that stuff and then we have a `service/crt` service package with its own `Create` and other stuff. For now when having the cert-operator manage CAs and Certs, the `service/crt` service should import and use the `service/ca` service. Later we can decouple them more easily. WDYT?

service/create/service.go
```diff
@@ -94,24 +94,42 @@ func (s *Service) Boot() {
...
+		err := s.setupPKIBackend(cert)
 		if err == nil {
-			s.Config.Logger.Log("info", "Cert issued")
-			s.Config.Logger.Log("info", cert.CommonName)
-			s.Config.Logger.Log("info", issueResp.SerialNumber)
+
+			// Ensure a PKI policy exists for the cluster.
+			err := s.setupPKIPolicy(cert)
+			if err == nil {
+
```
xh3b4sd: Dat empty line. Are they on purpose? I think we don't need them. Being picky I know right.

main.go
rossf7: Cool. I'll watch for this in future and I've raised #12 to rename these flags.

service/create/pki.go
rossf7: I've opened #11 to cover splitting Create into 2 services. I agree its a cleaner design and I'll also look at the public / private methods problem.

rossf7:
@xh3b4sd I've fixed the comments and whitespace problems and raised issues for the other issues. So I think this is ready now.

service/create/pki.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	isValid, err := service.VerifyPKISetup(cert.ClusterID)
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+	if isValid {
+		s.Config.Logger.Log("debug", fmt.Sprintf("PKI backend already exists for cluster %s", cert.ClusterID))
+		return nil
+	} else {
```
JosephSalisbury: I feel like it's more idiomatic to remove the `else` here. e.g:

```
if isValid {
   s.Config.Logger....
   return nil
}

caCommonName := s.Con...
```

service/create/pki.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	isCreated, err := service.IsPolicyCreated(cert.ClusterID)
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+	if isCreated {
+		s.Config.Logger.Log("debug", fmt.Sprintf("PKI policy already exists for cluster %s", cert.ClusterID))
+		return nil
+	} else {
```
JosephSalisbury: Here as well

rossf7:
@JosephSalisbury I've removed both elses to make the error handling more idiomatic.

service/create/service.go
```diff
@@ -94,24 +94,40 @@ func (s *Service) Boot() {
...
-		issueResp, err := s.Issue(cert)
+		// Ensure a PKI backend exists for the cluster.
+		err := s.setupPKIBackend(cert)
 		if err == nil {
-			s.Config.Logger.Log("info", "Cert issued")
-			s.Config.Logger.Log("info", cert.CommonName)
-			s.Config.Logger.Log("info", issueResp.SerialNumber)
+			// Ensure a PKI policy exists for the cluster.
+			err := s.setupPKIPolicy(cert)
+			if err == nil {
```
xh3b4sd: I noticed that this here is `err == nil`. It is usually `err != nil`. So this can easily be overlooked. I would like to agree on one way to handle error conditions to make our lives more easy. I think `err != nil` is more common. Lets create an issue for this and improve it separately.

service/create/service.go
```diff
@@ -94,24 +94,40 @@ func (s *Service) Boot() {
...
+				"kubernetes.default",
+				"kubernetes.default.svc",
+				"kubernetes.default.svc.cluster.local",
+			},
+			AllowBareDomains: true,
 			TTL:              "720h",
 		}
 
-		issueResp, err := s.Issue(cert)
+		// Ensure a PKI backend exists for the cluster.
```
JosephSalisbury:
```
if err := s.setupPKIBackend(cert); err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not setup pki backend '%#v'", err))
   return
}

if err := s.setupPKIPolicy(cert); err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not setup pki policy '%#v'", err))
   return
}

issueResp, err := s.Issue(cert)
if err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not issue cert '%#v'", err))
   return
}
s.Config.Logger.Log("info", fmt.Sprintf("cert issue %s %s", cert.CommonName, issueResp.SerialNumber))
```

does this make sense (it's untested :D) ? i'd say this is more idiomatic. up for discussion, though

JosephSalisbury: I'm not a fan of how nested this is, i found it hard to parse

xh3b4sd: I like that more as well, but lets maybe get this PR merged so we can have these things sorted out in separate PRs.

JosephSalisbury: Works for me 👍

JosephSalisbury: plz make sure there's an issue mr @rossf7

rossf7: Yes I'd like to get this merged but I've added this to #13 as I agree its nicer :)

Issue #13, opened 2017-03-23
cert-operator: check error handling follows standard conventions


rossf7:
Watch for non standard error handling e.g. if err == nil should be if err != nil

Follow Go idioms wherever possible.

https://github.com/giantswarm/cert-operator/blob/setup-pki/service/create/service.go#L116

rossf7:
Also simplify PKI setup error handling when integrating the watch. This from @JosephSalisbury would be cleaner.

```
if err := s.setupPKIBackend(cert); err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not setup pki backend '%#v'", err))
   return
}

if err := s.setupPKIPolicy(cert); err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not setup pki policy '%#v'", err))
   return
}

issueResp, err := s.Issue(cert)
if err != nil {
   s.Config.Logger.Log("error", fmt.Sprintf("could not issue cert '%#v'", err))
   return
}
s.Config.Logger.Log("info", fmt.Sprintf("cert issue %s %s", cert.CommonName, issueResp.SerialNumber))
```

PR #14, opened 2017-03-23
Integrate certificatetpr into operator

rossf7:
Towards giantswarm/giantswarm#1255 and #13 

This PR integrates the certificatetpr, creates the TPR on boot and watches for cert TPOs. The AddFunc is implemented and will issue a cert. The error handling for issuing certs has been cleaned up as requested in #13

Still to be implemented

- DeleteFunc
- UpdateFunc
- Saving certs as k8s secrets
- Retry logic on failure

I'd like to do these as follow on PRs.

~~I'm also still tracking down a bug with bare domains not being accepted as alternate names for the certs.~~

### DIFF
--- a/service/create/service.go
+++ b/service/create/service.go
@@ -14,16 +14,6 @@ import (
 	"github.com/giantswarm/cert-operator/flag"
 )
 
-// TODO Replace with Certificate TPR
-type CertificateSpec struct {
-	ClusterID        string
-	CommonName       string
-	IPSANs           []string
-	AltNames         []string
-	AllowBareDomains bool
-	TTL              string
-}
-
 // Config represents the configuration used to create a create service.
 type Config struct {
 	// Dependencies.
@@ -98,7 +88,7 @@ func (s *Service) Boot() {
 		}
 		s.Config.Logger.Log("info", "successfully created third-party resource")
 
-		cert := CertificateSpec{
+		cert := certificatetpr.Spec{
 			ClusterID:  "cert-test",
 			CommonName: "api.cert-test.g8s.eu-west-1.aws.test.private.giantswarm.io",
 			IPSANs:     []string{"10.0.0.4", "10.0.0.5"},
--- a/service/create/service.go
+++ b/service/create/service.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/giantswarm/certificatetpr"
 	microerror "github.com/giantswarm/microkit/error"
 	micrologger "github.com/giantswarm/microkit/logger"
 	vaultapi "github.com/hashicorp/vault/api"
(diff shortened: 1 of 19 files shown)


service/create/third_party_resource.go
```diff
@@ -0,0 +1,27 @@
...
+		ObjectMeta: v1.ObjectMeta{
+			Name: certificatetpr.Name,
+		},
+		Versions: []v1beta1.APIVersion{
+			{Name: "v1"},
+		},
+		Description: "Managed certificates on Kubernetes clusters",
+	}
+	_, err := s.Config.K8sClient.Extensions().ThirdPartyResources().Create(tpr)
+	if err != nil && !errors.IsAlreadyExists(err) {
```
xh3b4sd: I would find it cool if the conditionals would be more declarative and separated like this. It is easier for my weak brain to read it. 

```
if errors.AlreadyExists(err) {
    return nil
} else if err != nil {
    return microerror.MaskAny(err)
}
```

service/create/service.go
```diff
@@ -89,45 +96,101 @@ type Service struct {
...
+			panic(fmt.Sprintf("could not create cluster resource: %#v", err))
 		}
+		s.Config.Logger.Log("info", "successfully created third-party resource")
+
+		_, certInformer := cache.NewInformer(
+			s.newCertificateListWatch(),
+			&certificatetpr.CustomObject{},
+			resyncPeriod,
+			cache.ResourceEventHandlerFuncs{
+				AddFunc:    s.handleCertAdd,
```
xh3b4sd: The relations would be more easy to map if we align the function names with the struct properties IMO. So `handleCertAdd` would become `addFunc`.

service/create/service.go
```diff
@@ -89,45 +96,101 @@ type Service struct {
...
+}
+
+// handleCertDelete is not yet implemented.
+func (s *Service) handleCertDelete(obj interface{}) {
+	cert := obj.(*certificatetpr.CustomObject)
+	s.Config.Logger.Log("info", fmt.Sprintf("deleting certificate '%s' is not implemented yet", cert.Spec.CommonName))
+}
+
+// handleCertUpdate is not yet implemented.
+func (s *Service) handleCertUpdate(old, cur interface{}) {
```
JosephSalisbury: It might make life easier to try and formalise your signatures earlier, so like, I would kind of expect this to return `error` or something, to save you some time in the future. No blocker or anything.

rossf7:
@xh3b4sd good nits I've implemented 👍 

@JosephSalisbury the event handlers don't seem to return an error. The Prometheus [operator](https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L249) operator uses a queue here. Maybe for error handling? I'll add to #13 as I agree I think we need to worry about this.

JosephSalisbury:
Oh, maybe it doesn't make sense for the event handlers, as the error wouldn't really go anyhere. Ignore me, i wasn't considering the context

PR #15, opened 2017-03-24
Save certificates as k8s secrets and implement delete watch

rossf7:

This PR implements saving the certs issued by Vault as a k8s secret. It also implements the delete watch which deletes the secret when the TPO is deleted.

Not yet implemented is updateFunc as well as retry logic for Vault and K8s API calls. These will be done as follow up PRs.

### DIFF
--- a/service/create/k8s_secret.go
+++ b/service/create/k8s_secret.go
@@ -39,3 +39,16 @@ func (s *Service) SaveCertificate(cert certificateSecret) error {
 
 	return nil
 }
+
+// DeleteCertificate deletes the k8s secret that stores the certificate.
+func (s *Service) DeleteCertificate(cert *certificatetpr.CustomObject) error {
+	namespace := cert.ObjectMeta.Namespace
+	secretName := cert.Spec.CommonName
+
+	err := s.Config.K8sClient.Core().Secrets(namespace).Delete(secretName, &v1.DeleteOptions{})
+	if err != nil {
+		return microerror.MaskAny(err)
+	}
+
+	return nil
+}
--- a/service/create/k8s_secret.go
+++ b/service/create/k8s_secret.go
@@ -14,6 +14,9 @@ func (s *Service) CreateCertificate(cert certificateSecret) error {
 	secret := &v1.Secret{
 		ObjectMeta: v1.ObjectMeta{
 			Name: cert.CommonName,
+			Labels: map[string]string{
+				"clusterComponent": cert.ClusterComponent,
+			},
 		},
 		StringData: map[string]string{
 			"crt": cert.IssueResponse.Certificate,
--- a/service/create/k8s_secret.go
+++ b/service/create/k8s_secret.go
@@ -1,0 +1,41 @@
+package create
+
+import (
+	"github.com/giantswarm/certificatetpr"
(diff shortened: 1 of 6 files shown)


service/create/k8s_secret.go
```diff
@@ -0,0 +1,54 @@
...
+			"key": cert.IssueResponse.PrivateKey,
+			"ca":  cert.IssueResponse.IssuingCA,
+		},
+	}
+
+	// Create the secret.
+	_, err = s.Config.K8sClient.Core().Secrets(cert.Namespace).Create(secret)
+	if errors.IsAlreadyExists(err) {
+		// Update the secret if it already exists.
+		_, err = s.Config.K8sClient.Core().Secrets(cert.Namespace).Update(secret)
```
rossf7: If the secret already exists then I'm saving the updated cert. But I'm not sure this is right? It could also be an error state as it shouldn't happen if the watch is working correctly.

service/create/service.go
```diff
@@ -42,6 +43,13 @@ type Config struct {
 	Viper *viper.Viper
 }
 
+// certificateSecret stores a cert issued by Vault that will be stored as a k8s secret.
```
rossf7: I made this private but maybe it should be public if we want to expose issue cert as a library?

service/create/k8s_secret.go
```diff
@@ -0,0 +1,54 @@
...
+		if err != nil {
+			return microerror.MaskAny(err)
+		}
+
+		return nil
+
```
xh3b4sd: Dat empty line.

service/create/k8s_secret.go
xh3b4sd: I feel like if the certificate already exists `SaveCertificate` should ignore it to be idempotent or return an error. We should then have an `UpdateCertificate` method to update some existing data, which fails in case there does not data exist yet. WDYT?

service/create/k8s_secret.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+// DeleteCertificate deletes the k8s secret that stores the certificate.
+func (s *Service) DeleteCertificate(cert *certificatetpr.CustomObject) error {
+	namespace := cert.ObjectMeta.Namespace
+	secretName := cert.Spec.CommonName
+
+	err := s.Config.K8sClient.Core().Secrets(namespace).Delete(secretName, &v1.DeleteOptions{})
+	if err != nil {
```
xh3b4sd: I think delete operations should always be idempotent. WDYT?

service/create/service.go
xh3b4sd: Uncertain. We can iterate on this.

service/create/k8s_secret.go
rossf7: Yes I think that's better. I think UpdateCertificate should be triggered by the update watch. I'll change the create logic to make it idempotent.

service/create/k8s_secret.go
rossf7: Agreed I'll change it.

rossf7:
@xh3b4sd could you take another look?

Creating and deleting the secrets are now idempotent. Also the secret is labelled with the cluster component from the TPO.

PR #18, opened 2017-03-28
Add example certs for all cluster components.

rossf7:

This PR adds example certs for each cluster component. The actual certificate TPOs will be created by kubernetesd using data from the cluster TPO but there are some gaps currently.

- apiserver
- worker
- etcd
- calico

I'd like to get these TPOs as close to the actual ones as possible. I'll then automate in kubernetesd. For now I've agreed with @asymmetric that we'll use the default namespace. This is while we decide if the TPOs should live in the cluster namespace.

### DIFF
--- a/examples/apiserver-cert.yaml
+++ b/examples/apiserver-cert.yaml
@@ -1,20 +1,20 @@
 apiVersion: "giantswarm.io/v1"
 kind: Certificate
 metadata:
-  name: "example-cert"
+  name: "uo91f-apiserver"
+  namespace: "uo91f"
 spec:
-  clusterID: "cert-test"
-  clusterComponent: "kubernetes"
-  commonName: "api.cert-test.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  clusterID: "uo91f"
+  clusterComponent: "apiserver"
+  commonName: "api.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
   altNames:
-  - "cert-test.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  - "uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
   - "k8s-master-vm"
   - "kubernetes"
   - "kubernetes.default"
   - "kubernetes.default.svc"
   - "kubernetes.default.svc.cluster.local"
   ipSans:
-  - "10.0.2.1"
   - "10.0.2.2"
   ttl: "720h"
   allowBareDomains: true
--- a/examples/apiserver-cert.yaml
+++ b/examples/apiserver-cert.yaml
@@ -2,7 +2,7 @@ apiVersion: "giantswarm.io/v1"
 kind: Certificate
 metadata:
   name: "uo91f-apiserver"
-  namespace: "uo91f"
+  namespace: "default"
 spec:
   clusterID: "uo91f"
   clusterComponent: "apiserver"
--- a/examples/calico-cert.yaml
+++ b/examples/calico-cert.yaml
@@ -1,0 +1,14 @@
+apiVersion: "giantswarm.io/v1"
+kind: Certificate
+metadata:
+  name: "uo91f-calico"
+  namespace: "uo91f"
+spec:
+  clusterID: "uo91f"
+  clusterComponent: "calico"
+  commonName: "calico.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  altNames:
+  ipSans:
+  - "10.0.2.2"
+  ttl: "720h"
+  allowBareDomains: false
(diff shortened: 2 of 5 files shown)


examples/apiserver-cert.yaml
```diff
@@ -1,20 +1,20 @@
 apiVersion: "giantswarm.io/v1"
 kind: Certificate
 metadata:
-  name: "example-cert"
+  name: "uo91f-apiserver"
+  namespace: "uo91f"
```
rossf7: Should the TPOs be in the "per cluster" or the giantswarm namespace?

examples/calico-cert.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: "giantswarm.io/v1"
+kind: Certificate
+metadata:
+  name: "uo91f-calico"
+  namespace: "uo91f"
+spec:
+  clusterID: "uo91f"
+  clusterComponent: "calico"
+  commonName: "calico.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
```
rossf7: Is the calico domain correct? It isn't in the clustertpr, should we add it?

https://github.com/giantswarm/clustertpr/blob/master/calico/calico.go

examples/worker-cert.yaml
```diff
@@ -0,0 +1,20 @@
...
+metadata:
+  name: "uo91f-worker"
+  namespace: "uo91f"
+spec:
+  clusterID: "uo91f"
+  clusterComponent: "worker"
+  commonName: "api.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  altNames:
+  - "uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  - "k8s-master-vm"
```
puja108: why does the worker cert need these alt names? these are the api server ANs aren't they?

puja108:
There's 2 sides to the TPO namespace issue:
1. I was thinking that TPO could be independent of namespace and either kubernetesd or some operator creates the namespace based on it.
2. Thinking about it a bit more the TPO actually belongs to the cluster just like the deployments and secrets to and having everythin gpertaining to a single cluster live in a separate namespace would help with order and cleaning up (e.g. cleaning up after a deleted cluster is just deleting the whole namespace).

So now I'm a bit more tending toward 2

examples/worker-cert.yaml
rossf7: I think that's an error on my side, I'll remove them. I'd like to get these examples as close as possible and then I can automate in kubernetesd.

examples/calico-cert.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: "giantswarm.io/v1"
+kind: Certificate
+metadata:
+  name: "uo91f-calico"
+  namespace: "default"
+spec:
+  clusterID: "uo91f"
+  clusterComponent: "calico"
+  commonName: "calico.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
```
rossf7: @xh3b4sd the Calico domain is missing from clustertpr. Can we add it and is the example correct?

xh3b4sd: See 

- https://github.com/giantswarm/giantswarm/issues/1305
- https://github.com/giantswarm/giantswarm/issues/1304

Feel free to do that in case your are blocked by it somehow. The domains for the guest cluster components have all the same structure and only differ in their sub domain. I don't know how the base domain of AWS clusters look like. And note that the kube DNS domain is a completely different thing structure wise.

rossf7: @xh3b4sd will do if it blocks me. Also sorry about forgetting my own issues!

examples/worker-cert.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: "giantswarm.io/v1"
+kind: Certificate
+metadata:
+  name: "uo91f-worker"
+  namespace: "default"
+spec:
+  clusterID: "uo91f"
+  clusterComponent: "worker"
+  commonName: "api.uo91f.g8s.eu-west-1.aws.test.private.giantswarm.io"
```
xh3b4sd: Common name for the worker is `worker`, not `api`.

examples/apiserver-cert.yaml
```diff
@@ -1,20 +1,20 @@
...
 metadata:
-  name: "example-cert"
+  name: "uo91f-apiserver"
+  namespace: "default"
 spec:
-  clusterID: "cert-test"
-  clusterComponent: "kubernetes"
-  commonName: "api.cert-test.g8s.eu-west-1.aws.test.private.giantswarm.io"
+  clusterID: "uo91f"
+  clusterComponent: "apiserver"
```
xh3b4sd: I would prefer `api` over `apiserver`. The scheme for the other certs is to use the `clusterComponent` as sub domain identifier. The only exception is here.

examples/worker-cert.yaml
rossf7: Thanks I'll fix.

PR #19, opened 2017-03-28
Naming changes for k8s secrets created by cert-operator

rossf7:
Toward giantswarm/giantswarm#1255

This PR implements changes to the k8s secrets created by the operator so it can integrate with the AWS and KVM operators.

- Secrets are now labelled with the clusterID and clusterComponent so they can be selected by the operators.
- Magic strings have been replaced with types in certificatetpr.


service/crt/k8s_secret.go
```diff
@@ -51,3 +51,7 @@ func (s *Service) DeleteCertificate(cert *certificatetpr.CustomObject) error {
 
 	return nil
 }
+
+func (s *Service) getSecretName(cert *certificatetpr.CustomObject) string {
```
asymmetric: I would have this as a simple function, since it doesn't need to be a method. Not sure what the most idiomatic Go way is though.

service/crt/k8s_secret.go
```diff
@@ -1,32 +1,35 @@
...
-				"clusterComponent": cert.ClusterComponent,
+				certificatetpr.ClusterIDLabel: secret.Certificate.Spec.ClusterID,
+				certificatetpr.ComponentLabel: secret.Certificate.Spec.ClusterComponent,
 			},
 		},
 		StringData: map[string]string{
-			"crt": cert.IssueResponse.Certificate,
-			"key": cert.IssueResponse.PrivateKey,
-			"ca":  cert.IssueResponse.IssuingCA,
+			certificatetpr.Crt.String(): secret.IssueResponse.Certificate,
```
asymmetric: Does it work if you don't add `.String()`?

service/crt/k8s_secret.go
rossf7: Good point, I'll change this to a function.

service/crt/k8s_secret.go
rossf7: Without the .String() it errors because certificatetpr.TLSAssetType is a type. Its a bit ugly but I think the .String() is needed.

PR #23, opened 2017-04-06
Updated tpos to match awstpr example

rossf7:
Updated TPOs for testing to match awstpr example.

Can we confirm that api.example-cluster.g8s.eu-west-1.aws.private.giantswarm.io is the correct format for the CN?

### DIFF
--- a/examples/api-cert.yaml
+++ b/examples/api-cert.yaml
@@ -1,14 +1,14 @@
 apiVersion: "giantswarm.io/v1"
 kind: Certificate
 metadata:
-  name: "uo91f-api"
+  name: "example-cluster-api"
   namespace: "default"
 spec:
-  clusterID: "uo91f"
+  clusterID: "example-cluster"
   clusterComponent: "api"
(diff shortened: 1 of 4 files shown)


asymmetric:
What I know is that the ELB in front of the masters has a public DNS name of *example-cluster.aws.giantswarm.io*.

xh3b4sd:
@rossf7 what about this PR? Is quite old already. Can we merge? Should we close it?

rossf7:
Closing as the current examples are fine for testing. I wanted to make the examples match the AWS setup exactly but I no longer think this is necessary.

asymmetric:
I think the CN is wrong. The Domain name is like `api.example-cluster.aws.giantswarm.io`.

rossf7:
@asymmetric the format is api.example-cluster.g8s.eu-west-1.[customer].aws.giantswarm.io 

This is set in the architect installation and is templated in kubernetesd https://github.com/giantswarm/architect/blob/master/installation/aws.go#L75

PR #38, opened 2017-04-27
Use multiple args for config file flags

rossf7:
Towards giantswarm/giantswarm#1388

Fixes problem with operator failing to start because values can't be read from the config map. This is because the config args are being parsed as strings rather than string slices. Meaning that the secrets file is read but not the config map.

The fix is to specify the config files and dirs examples multiple times as shown in this pflag [issue](https://github.com/spf13/pflag/issues/40). The bug is strange because it occurs in k8s but not locally.

### DIFF
--- a/kubernetes/deployment.yaml
+++ b/kubernetes/deployment.yaml
@@ -50,8 +50,10 @@ spec:
           containerPort: 8000
         args:
         - daemon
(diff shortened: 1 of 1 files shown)


rossf7:
Thanks <3 Yes lets coordinate and get this fixed everywhere else.

rossf7:
I'll get the microkit test finished too as its nearly there and will be useful.

PR #43, opened 2017-05-16
Docs for running cert-operator locally

### LINKED ISSUE #40: Docs for running cert-operator locally

rossf7:
Fixes #40 

Adds docs for running the operator locally. Also helps with running aws-operator locally as the example cert TPOs match the example cluster TPO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,12 @@ the Kubernetes API. For development running Vault in dev mode is fine.
 ### Setup
 
 - The operator needs to connect to a Vault server. See this [example config](https://gist.github.com/rossf7/106fc1c97ebf24517d8be3cb30eb9a49) for running Vault as a deployment with a ClusterIP service.
-- The cert-operator binary needs to be built into a docker image and tagged as
-`quay.io/giantswarm/cert-operator:local-dev`.
+- The cert-operator binary needs to be built into a docker image and tagged as `quay.io/giantswarm/cert-operator:local-dev`. The current pod need to be deleted for changes to apply.
 
 ```
-GOOS=linux go build github.com/giantswarm/cert-operator && docker build -t quay.io/giantswarm/cert-operator:local-dev .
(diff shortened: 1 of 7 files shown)


kopiczko:
@rossf7 why not include vault.yml in examples?

rossf7:
@kopiczko can you delete the operator pod and try again? I've seen that problem the first time it runs with creating the TPR.

On vault.yml being in /examples/ I thought it was simpler as a gist but I don't have a strong opinion. Currently we have to use Vault 0.6.4 due to `certctl` but we'll need to upgrade to Vault 0.7 for the Kubernetes 1.6 upgrade.

kopiczko:
@rossf7 
1. I have to delete pod every time after deployment on a fresh cluster, bc of `Failed to list *certificatetpr.CustomObject`
2. I lost some time on realising I need to replace YOUR_VAULT_HOST. While obvious now it may save some time other folks if this is stated in the readme.

rossf7:
@kopiczko sorry I've added a note on editing YOUR_VAULT_HOST. I also added the delete pod step so the new docker image is used.

I'll look into the startup bug with the TPR.

asymmetric:
I agree that the Vault config should be in `examples/`.

kopiczko:
Maybe create `local-vault.yml` and `local-deployment.yml` in examples  & replace YOUR_VAULT_HOST with `vault` as service is in place.

That'd be our default way to set local cert-operator.

Let me know your thoughts.

rossf7:
@kopiczko the problem with using the service hostname is then you also need `kube-dns` running on your local cluster. So it adds another dependency.

This is a simple setup for local testing. For pull based deployments hopefully it will support local targets. That would be useful for local testing and we could stop using this manual setup.

cc @JosephSalisbury

rossf7:
I've created #44 for the startup error.

Issue #32, opened 2017-04-21
cert-operator should delete secrets when the certs TPO is deleted


rossf7:
- AWS integration testing has found a problem with secrets not being deleting.
- With a large number of secrets aws-operator has problems reading them.

Need to investigate and fix as the delete watch should do this.

rossf7:
When deployed to AWS the cert-operator fails to delete the k8s secrets it generates and this decode error appears in the logs. The error happens before the delete logic runs which is why the k8s secrets are not deleted.

```
streamwatcher.go:109] Unable to decode an event from the watch stream: json: cannot unmarshal string into Go value of type struct { Type watch.EventType; Object certificatetpr.CustomObject }
```

Searching for this error leads to this [k8s](https://github.com/kubernetes/kubernetes/issues/26003) issue and this [client-go](https://github.com/kubernetes/client-go/issues/8) issue. Based on these I want to update certificatetpr to match the [example](https://github.com/kubernetes/client-go/blob/release-2.0/examples/third-party-resources/types.go) in the v2.0.0 branch. This includes a workaround for known issues with TPRs and ugorji decoding.

rossf7:
Did more digging yesterday and this [Traefik issue](https://github.com/containous/traefik/issues/532) looks to be the same problem. From the issue and the [PR](https://github.com/containous/traefik/pull/836) they fixed it by setting a resync period of 5 minutes. 

Currently we set this to 0 which delays the resync for as long as possible.
https://github.com/giantswarm/cert-operator/blob/master/service/crt/service.go#L33

I'm going to extract the TPR watch to `operatorkit` but I'd like to try this change in `cert-operator` first.

PR #47, opened 2017-05-17
Set resync period to 5 minutes to fix problem with delete watch

rossf7:
See the Traefik issue and PR linked in #32

This is an attempt to fix why delete watches aren't being triggered. Sets resync period to 5 minutes. Currently this is set to 0 so it delays the resync for as long as possible.

If this fix works we need to implement in operatorkit. As this problem also happens with `aws-operator` and means we need to clean up AWS resources manually.


rossf7:
I matched the Traefik change. If this works lets try reducing it. I can't replicate the delete watch problem locally so I want to test it on g8s aws.

PR #48, opened 2017-05-18
Revert resend period change it didn’t fix issue and made logs noisy

rossf7:
This change seems to have introduced another problem: the time it takes for the operator to create secrets has increased.


asymmetric:
Another option would be increasing the time-out on the `aws-operator`, since the certs do eventually get created. Unsure.

rossf7:
@asymmetric thanks for fixing the desc. I hadn't realised this was blocking launching clusters. I'm going to revert anyway due to the higher volume of logs.

But I think increasing the timeouts for the secrets watch could be a good idea too. Not sure what the current value is but this has happened a few times.

Issue #49, opened 2017-05-18
Reconfigure `resyncPeriod`


asymmetric:
We reverted this value in #48, but having a timeout actually makes sense.

The CoreOS people say they usually set it to 10 minutes, FWIW.

rossf7:
@asymmetric thanks for raising and totally agree we need a value for this. 0 is just wrong for a production operator.

The problem is the value to use as it varies a lot, including at CoreOS! ;) The Prometheus operator and Traefik use 5 minutes but the Rook operator uses 15 seconds. How about 1 minute? 

@xh3b4sd WDYT?

https://github.com/coreos/prometheus-operator/search?utf8=%E2%9C%93&q=resyncPeriod&type=
https://github.com/containous/traefik/search?utf8=%E2%9C%93&q=resyncPeriod&type=
https://github.com/rook/rook/search?utf8=%E2%9C%93&q=resyncPeriod&type=

xh3b4sd:
I don't know. We have to try it out and test what works for us.

PR #50, opened 2017-05-19
WIP: Delete watch fixes

rossf7:
This PR has the changes needed to get the delete watch working. I can only repro on AWS so needed to test on prod :(

Still trying a couple of things to see if all the changes are necessary.


service/crt/service.go
```diff
@@ -30,7 +29,7 @@ const (
 	// resyncPeriod is the period for re-synchronizing the list of objects in k8s
 	// watcher. 0 means that re-sync will be delayed as long as possible, until
 	// the watch will be closed or timed out.
-	resyncPeriod time.Duration = 0
+	resyncPeriod time.Duration = time.Minute * 10
```
rossf7: We need to set a value here. 0 is just wrong. See https://github.com/giantswarm/cert-operator/issues/49

service/crt/service.go
```diff
@@ -194,8 +193,9 @@ func (s *Service) newCertificateListWatch() *cache.ListWatch {
 				return nil, err
 			}
 
-			watcher := watch.NewStreamWatcher(&k8sutil.CertificateDecoder{
-				Stream: stream,
+			watcher := watch.NewStreamWatcher(&certificateDecoder{
+				decoder: json.NewDecoder(stream),
+				close:   stream.Close,
```
rossf7: The decoder here and in `aws-operator` is missing this close func which is definitely needed. To get it working I just moved the decoder into the service.

@xh3b4sd this was based on the decoder in `kvm-operator` which I love <3 Its also in `operatorkit` so we can fix it by using the code there.

vendor/github.com/giantswarm/certificatetpr/custom_object.go
```diff
@@ -12,7 +14,7 @@ import (
 // specifications of the resource the Certificate operator is interested in.
 type CustomObject struct {
 	unversioned.TypeMeta `json:",inline"`
-	Metadata             api.ObjectMeta `json:"metadata"`
+	Metadata             api.ObjectMeta `json:"metadata" yaml:"metadata"`
```
rossf7: I messed up when I applied the "decoding magic". The yaml encoding was in spec but not in CustomObject and List. I think this was making the decoder sad.

Its also possible we need to remove the yaml support entirely. Testing that next.

vendor/github.com/giantswarm/certificatetpr/custom_object.go
```diff
@@ -34,12 +36,19 @@ func (c *CustomObject) GetObjectMeta() meta.Object {
 type CustomObjectCopy CustomObject
 
 func (c *CustomObject) UnmarshalJSON(data []byte) error {
+	fmt.Fprintf(os.Stdout, "IN CustomObject UnmarshalJSON data: %s \n", data)
```
rossf7: This is the "decoding magic" from the k8s.io/client-go example. Plus the debug I hacked in.

See  https://github.com/kubernetes/client-go/blob/v2.0.0/examples/third-party-resources/types.go#L66

vendor/github.com/giantswarm/certificatetpr/list.go
```diff
@@ -31,12 +33,18 @@ func (l *List) GetListMeta() unversioned.List {
 type ListCopy List
 
 func (l *List) UnmarshalJSON(data []byte) error {
+	fmt.Fprintf(os.Stdout, "IN List UnmarshalJSON data: %s \n", data)
```
rossf7: Same thing is needed for the list.

PR #51, opened 2017-05-19
Fix delete watch problems

### LINKED ISSUE #32: cert-operator should delete secrets when the certs TPO is deleted

rossf7:
Fixes #32 and #49 

This PR fixes the problems with the delete watch. Previously when deleting a cert TPO it was removed but the DeleteFunc was not triggered so the secret was not deleted. Problems were

- certificatetpr has yaml attributes for the Spec but not the CustomObject and List.
- the decoder did not close the watch stream.

The decoder has been moved into the crt service for now. It will be replaced by `operatorkit` once the reconciliation loop support is moved there from `kvm-operator`.

This change also sets the resync period to 1 minute to make the watches more robust. See #49 multiple values are being used by other operators. We need to experiment and see what works for us.


service/crt/decoder.go
```diff
@@ -0,0 +1,38 @@
...
+	return e.Type, &e.Object, nil
+}
+
+// Close closes the stream.
+func (cd *certificateDecoder) Close() {
+	err := cd.close()
+	if err != nil {
+		panic(err)
+	}
+}
```
kopiczko: The only change I see here is that json.Decoder isn't created on every Decode invocation. Which makes a lot of sense. But the stream were closed by a previous Close implementation.

Please correct me if I'm wrong.

service/crt/service.go
```diff
@@ -193,8 +192,9 @@ func (s *Service) newCertificateListWatch() *cache.ListWatch {
 				return nil, err
 			}
 
-			watcher := watch.NewStreamWatcher(&k8sutil.CertificateDecoder{
-				Stream: stream,
+			watcher := watch.NewStreamWatcher(&certificateDecoder{
+				decoder: json.NewDecoder(stream),
+				close:   stream.Close,
```
kopiczko: Why not init json.Decoder in newCertificateDecoder(stream)?

service/crt/service.go
```diff
@@ -28,9 +27,8 @@ const (
 	CertificateWatchAPIEndpoint string = "/apis/giantswarm.io/v1/watch/certificates"
 
 	// resyncPeriod is the period for re-synchronizing the list of objects in k8s
-	// watcher. 0 means that re-sync will be delayed as long as possible, until
-	// the watch will be closed or timed out.
-	resyncPeriod time.Duration = 0
+	// watcher. Set to 1 minute to make the watch more robust.
+	resyncPeriod time.Duration = time.Minute * 1
```
xh3b4sd: Why not 30 seconds? Why not 1 second?

service/crt/decoder.go
rossf7: Not sure and some of this is controlled by the informer. I don't have a full picture of what the informer is doing under the hood and there are no docs.

@asymmetric has been looking at this and talking to CoreOS about it.

service/crt/service.go
rossf7: That is maybe a bit cleaner but I don't have a strong opinion on it. We should move to the operatorkit support once its ready.

service/crt/service.go
rossf7: AFAIK the resync is a corrective action so the watch stays healthy since its a long running process. So I'd say 1 second is too often. The lowest value I've seen is 15 seconds in the Rook operator.

I agree on needing to experiment on this. I could use a flag to make that easier?

This is the best link I can find on what the resync period is for. The current value of 0 which delays the resync for as long as possible feels wrong to me.

https://github.com/kubernetes/kubernetes/issues/26406

xh3b4sd: Just curious. All good. Fine with the hard coded version. Lets iterate and find out.

service/crt/decoder.go
kopiczko: In PR description:

> the decoder did not close the watch stream.

I think the only change that is meaningful here is reusing json.Decoder. Stream instance was closed in previous implementation.
I'd like to make that clear, that's why I'm suggesting creating constructor function, like

```
func newcertificateDecoder(stream io.ReadCloser) {
    return certificateDecoder struct {
         decoder:  decoder: json.NewDecoder(stream),
         stream: stream,
     }
}
```

rossf7: @kopiczko thanks thats nice. Done

service/crt/decoder.go
```diff
@@ -15,6 +16,14 @@ type certificateDecoder struct {
 	close   func() error
```
kopiczko: Nitpicking: we can have a stream there and close it explicitly. I don't see a benefit of hiding it behind a function. If for whatever reason you want to hide Read method, you can use io.Closer type here.

rossf7: Yes I get your point. We can look at that for the `operatorkit` version.

service/crt/decoder.go
asymmetric: Is  there a reason why we provide our own decoder? I see [from another project](https://github.com/zalando-incubator/postgres-operator/blob/d34273543ed9aec2ddf97a611315586c615127bd/pkg/controller/postgresql.go#L71-L78) that you can delegate that to the client library.

xh3b4sd: I actually don't know. Would be interesting if this works as well. I am not sure how this then works though, because the object has to be defined so the decoder knows in which structure to parse.

rossf7: I thought we had to provide our own decoder but looks like that isn't true. I'd like to raise an issue in `operatorkit` for this. If we don't need the decoder it simplifies the operatorkit interface in a nice way.

rossf7: @asymmetric if you want to try that for the `aws-operator` PR it would be great to know if that works. For now I'd like to go with this so we can get the DeleteFunc working.

PR #53, opened 2017-05-22
Wait for TPR to init

kopiczko:
Towards https://github.com/giantswarm/cert-operator/issues/44


service/crt/service.go
```diff
@@ -98,11 +105,22 @@ func New(config Config) (*Service, error) {
...
 
+	tprConfig := tpr.Config{
+		Clientset:   config.K8sClient,
+		Name:        TPRName,
+		Domain:      TPRDomain,
+		Version:     TPRVersion,
+		Description: TPRDescription,
+	}
+	tpr, err := tpr.New(tprConfig)
+	microerror.PanicOnError(err)
```
kopiczko: May be controversial, but there is no reason for tar to fail.

service/crt/service.go
```diff
@@ -29,6 +30,11 @@ const (
 	// resyncPeriod is the period for re-synchronizing the list of objects in k8s
 	// watcher. Set to 1 minute to make the watch more robust.
 	resyncPeriod time.Duration = time.Minute * 1
+
+	TPRName        = "certificate"
```
xh3b4sd: The TPR name should actually come from https://github.com/giantswarm/certificatetpr/blob/989414c916fafbc70d7e09fb48ef053b7290a9e4/name.go#L5, but there is the complete name. Is the interface of the abstraction in operatorkit wrong?

service/crt/service.go
xh3b4sd: I would like to dispatch the error to the caller. The calling function returns an error in the interface. IMO methods/functions providing such interfaces should never panic. The error should always be forwarded to the caller. He has then to decide what to do with it.

service/crt/service.go
rossf7: We already have the name in the TPR. I'd like to see this integrated rather than being duplicated here.

https://github.com/giantswarm/certificatetpr/blob/master/name.go

kopiczko: @xh3b4sd I don't think the abstraction is wrong. My thinking was to make it possible to create list/watch URLs from these values. While cert-operator uses certificatetpr for TPR it has watch/list URLs baked in. Which I find even worse bc it gives false feeling it's configureable.

xh3b4sd: Fine, then lets fix the TPR repos to provide these information.

service/crt/service.go
kopiczko: Done.

service/crt/service.go
kopiczko: https://github.com/giantswarm/giantswarm/issues/1483

kopiczko:
@rossf7 I know but:
1. It's first operator using TPR from operatorkit
2. It's mostly your code
3. I think I heard most of the time we should have >1 approvals?

So I thought we better agree on that :)

rossf7:
I'm not sure about #3 but yes makes sense for me to review due to 1 & 2. All good and I like the progress being made on `operatorkit` :)

JosephSalisbury:
Random drive by - you only need 1 approval to merge. Waiting for more than one is at your own discretion. You're both being too good :D

PR #56, opened 2017-05-30
Use latest operatorkit

kopiczko:
* glide: update deps

* check IsAlreadyExists

* use *Endpoint functions


service/crt/service.go
```diff
@@ -23,11 +23,6 @@ import (
 )
 
 const (
-	CertificateListAPIEndpoint  string = "/apis/giantswarm.io/v1/certificates"
-	CertificateWatchAPIEndpoint string = "/apis/giantswarm.io/v1/watch/certificates"
-
-	TPRName        = "certificate"
-	TPRDomain      = "giantswarm.io"
 	TPRVersion     = "v1"
 	TPRDescription = "Managed certificates on Kubernetes clusters"
```
xh3b4sd: This should be in the TPR repo. Lets track an issue for this in our boards in case we are not doing it right away.

service/crt/service.go
```diff
@@ -126,7 +120,11 @@ func New(config Config) (*Service, error) {
 // Boot starts the service and implements the watch for the certificate TPR.
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
-		if err := s.tpr.CreateAndWait(); err != nil {
+		err := s.tpr.CreateAndWait()
+		if err != nil && tpr.IsAlreadyExists(err) {
+			s.Config.Logger.Log("info", "certificate resource already exists")
+		}
+		if err != nil {
```
xh3b4sd: FYI I would prefer this. 

```
if tpr.IsAlreadyExists(err) { 
    //
} else if err != nil {
    //
}
```

service/crt/service.go
kopiczko: All tracked here: https://github.com/giantswarm/giantswarm/issues/1483

service/crt/service.go
```diff
@@ -126,10 +120,15 @@ func New(config Config) (*Service, error) {
...
-		if err := s.tpr.CreateAndWait(); err != nil {
+		err := s.tpr.CreateAndWait()
+		switch {
+		case tpr.IsAlreadyExists(err):
+			s.Config.Logger.Log("info", "certificate third-party resource already exists")
+		case err != nil:
 			panic(fmt.Sprintf("could not create certificate resource: %#v", err))
+		default:
+			s.Config.Logger.Log("info", "successfully created certificate third-party resource")
 		}
```
kopiczko: @xh3b4sd I changed if to the above.

My take on that is `err != nil && tpr.IsAlreadyExists(err)` protects from panics.

xh3b4sd: Why should it panic?

kopiczko: I should say "protects from potential panics", because of passing nil as an argument. If for example errgo.Cause implementation changed to not handle nil or sth. I know it probably won't be the case, but it feels right, yadda yadda yadda :D

xh3b4sd: In this case all of our code would be wrong.

PR #58, opened 2017-06-21
Adds retry support to methods called by the operator

### LINKED ISSUE #10: cert-operator needs retry logic if creating the cert fails
Ideally should use exponential backoff retry logic from microkit.

rossf7:
Fixes #10 

This is some boyscouting I did a while back I want to finish off. It adds retry support to the methods called in the addFunc and deleteFunc.

### DIFF
--- a/service/ca/pki.go
+++ b/service/ca/pki.go
@@ -27,9 +27,10 @@ func (s *Service) SetupPKIAndWait(cert certificatetpr.Spec) error {
 		err := s.SetupPKI(cert)
 		if err != nil {
 			s.Logger.Log("info", "failed to setup PKI - retrying")
+			return microerror.MaskAny(err)
 		}
 
-		return err
+		return nil
 	}
 
 	return backoff.Retry(operation, initBackoff)
--- a/service/ca/pki.go
+++ b/service/ca/pki.go
@@ -3,14 +3,39 @@ package ca
 import (
 	"fmt"
 	"strings"
+	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/certctl/service/pki"
 	"github.com/giantswarm/certctl/service/token"
 	"github.com/giantswarm/certificatetpr"
 	microerror "github.com/giantswarm/microkit/error"
 )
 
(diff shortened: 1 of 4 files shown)


service/ca/pki.go
```diff
@@ -3,14 +3,39 @@ package ca
...
+	"github.com/cenkalti/backoff"
 	"github.com/giantswarm/certctl/service/pki"
 	"github.com/giantswarm/certctl/service/token"
 	"github.com/giantswarm/certificatetpr"
 	microerror "github.com/giantswarm/microkit/error"
 )
 
-// SetupPKI creates a PKI backend and policy if one does not exists for the cluster.
+const (
+	setupPKIMaxElapsedTime = 30 * time.Second
```
rossf7: The timeouts are short because the timeout for the secrets watch is only 90 seconds. The operator needs to issue all certs in that time.

service/ca/pki.go
```diff
@@ -3,14 +3,39 @@ package ca
...
+// the cluster. If an error occurs an exponential backoff is used. After the
+// max elapsed time the error is returned to the caller.
+func (s *Service) SetupPKIAndWait(cert certificatetpr.Spec) error {
+	initBackoff := backoff.NewExponentialBackOff()
+	initBackoff.MaxElapsedTime = setupPKIMaxElapsedTime
+
+	operation := func() error {
+		err := s.SetupPKI(cert)
+		if err != nil {
+			s.Logger.Log("info", "failed to setup PKI - retrying")
```
xh3b4sd: This error handling is a bit uncommon. I would prefer returning the error here masked like we usually do and do logging additionally if desired. Below should be returned nil. Looks strange that the error is always returned. In case `SetupPKI` does weird things we forward the weirdness here. Lets not do this.

rossf7: My thinking was that SetupPKI would have the normal logging. In the wait wrapper I just wanted to indicate that a retry was taking place.

But it is different to our normal error handling and I get the point about returning the error. I'll take another look.

rossf7: Done

PR #59, opened 2017-06-22
Add healthz endpoint that checks the vault token

rossf7:
Fixes giantswarm/giantswarm#1558

Adds healthz endpoint that checks the Vault token and API connection are working. To do this it lists the sys mounts. A health endpoint has been added to the Vault API and will be in the next release. We should switch to that once it's released and we've upgraded.

https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#next-unreleased


server/endpoint/endpoint.go
```diff
@@ -32,6 +33,18 @@ func DefaultConfig() Config {
 func New(config Config) (*Endpoint, error) {
 	var err error
 
+	var healthzEndpoint *healthz.Endpoint
```
rossf7: The endpoint is mainly boilerplate.

service/healthz/service.go
```diff
@@ -0,0 +1,97 @@
...
+	logger      micrologger.Logger
+
+	// Internals.
+	bootOnce sync.Once
+}
+
+// Check implements the health check which lists the mounts for the sys backend.
+// This checks that the operator can connect to the Vault API and the token is
+// valid.
+func (s *Service) Check(ctx context.Context, request Request) (*Response, error) {
```
rossf7: Check method is where the custom logic is. I chose to list the sys mounts as the API call as it shouldn't have any negative effects on the Vault server while still testing the token.

As mentioned in the PR desc I'd like to switch the new Vault health point once its released and we are up to date.

https://github.com/hashicorp/vault/blob/master/api/sys_health.go

service/healthz/service.go
```diff
@@ -0,0 +1,97 @@
+package healthz
```
rossf7: Service is also mainly boilerplate apart from the Check method.

service/healthz/service.go
```diff
@@ -0,0 +1,97 @@
...
+	"time"
+
+	microerror "github.com/giantswarm/microkit/error"
+	micrologger "github.com/giantswarm/microkit/logger"
+	vaultapi "github.com/hashicorp/vault/api"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	healthCheckRequests = prometheus.NewCounterVec(
```
rossf7: I based the Prometheus config on the healthz endpoint in kvm-operator.

kubernetes/deployment.yaml
```diff
@@ -54,6 +54,12 @@ spec:
         - --config.dirs=/var/run/cert-operator/secret/
         - --config.files=config
         - --config.files=secret
+        livenessProbe:
+          httpGet:
+            path: /
```
xh3b4sd: Why is this here `/` and in the Helm templates it is `/healthz`?

service/healthz/service.go
```diff
@@ -0,0 +1,97 @@
...
+	VaultClient *vaultapi.Client
+}
+
+// DefaultConfig provides a default configuration to create a new healthz
+// service by best effort.
+func DefaultConfig() Config {
+	return Config{
+		// Dependencies.
+		VaultClient: nil,
+		Logger:      nil,
```
xh3b4sd: Minor nitpick. Alphabetical would be awesome.

xh3b4sd:
Maybe also check kubernetesd. There is such a healthcheck already implemented. Maybe we also want the readiness probes and maybe some other differences are there.

rossf7:
I've fixed the /healthz path and param ordering. Thanks for the tip I'll check kubernetesd :)

service/healthz/service.go
```diff
@@ -0,0 +1,97 @@
...
+	bootOnce sync.Once
+}
+
+// Check implements the health check which lists the mounts for the sys backend.
+// This checks that the operator can connect to the Vault API and the token is
+// valid.
+func (s *Service) Check(ctx context.Context, request Request) (*Response, error) {
+	start := time.Now()
+	defer func() {
+		healthCheckRequestTime.Set(float64(time.Since(start) / time.Millisecond))
```
kopiczko: I think it should be `time.Millisecond - float64(time.Since(start)`

kopiczko: No this is right. Missed `Since`.

rossf7:
The liveness probe not working was a problem with my local minikube :( I deployed with a hacked circle and it was fine.

@xh3b4sd PTAL

helm/cert-operator-chart/templates/deployment.yaml
```diff
@@ -54,6 +54,12 @@ spec:
         - --config.dirs=/var/run/cert-operator/secret/
         - --config.files=config
         - --config.files=secret
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8000
+          initialDelaySeconds: 30
+          timeoutSeconds: 1
```
xh3b4sd: I think it would make sense to also define the readiness probe like in other deployments. Keeping the templates in sync will make it easier to unify or abstract them later, or at least to understand them.

server/server.go
```diff
@@ -114,21 +118,16 @@ func (s *server) Shutdown() {
...
-		default:
+		rErr := err.(microserver.ResponseError)
+		uErr := rErr.Underlying()
+
+		if rErr.IsEndpoint() {
+			json.NewEncoder(w).Encode(map[string]interface{}{
+				"code":    microserver.CodeFailure,
+				"message": uErr.Error(),
+				"from":    s.serviceName,
+			})
```
xh3b4sd: I was wondering why this is but I saw that this was also done this way in kubernetesd. This is in fact the default structure microkit already uses. See https://github.com/giantswarm/microkit/blob/1ac51fb9905e0c7ab5a8d12fec127c6fbfd75aac/server/server.go#L392-L394. So it is not necessary to do this. We probably implemented this during some transition period and forgot to clean it up. It should just be like this: https://github.com/giantswarm/api/blob/master/server/server.go#L236. You can set the error code and error message to the response error. This makes the code easier. Lets also make sure we clean this up in kubernetesd. You can go for it or I will do it later. However you like.

service/healthz/service.go
```diff
@@ -0,0 +1,111 @@
...
+	_, err := sysBackend.ListMounts()
+	if err != nil {
+		healthCheckRequests.WithLabelValues("failed").Inc()
+		return nil, microerror.MaskAny(err)
+	}
+
+	healthCheckRequests.WithLabelValues("successfull").Inc()
+
+	response := DefaultResponse()
+	response.Code = ResponseCodeOk
```
xh3b4sd: Same story here. Sorry I guided you through kubernetesd. We should use the standard codes IMO. See https://github.com/giantswarm/microkit/blob/master/server/codes.go#L24. This should also be fixed in kubernetesd. It is also that `ResponseCodeFailed` is not even used at all.

rossf7:
Closing in favour of #63

PR #60, opened 2017-06-25
Upgrade Vault to 0.7.3

rossf7:
Towards giantswarm/giantswarm#1366

This PR upgrades Vault from 0.6.4 to 0.7.3. Some minor code changes were required. I've tested issuing certs locally with minikube against both versions of Vault.


glide.yaml
```diff
@@ -5,7 +5,7 @@ import:
 - package: google.golang.org/grpc
   version: b5071124392bfd416c713e6595ac149b387f7186
 - package: github.com/giantswarm/certctl
-  version: master
+  version: vault-upgrade
```
rossf7: This needs to be changed to master

glide.yaml
```diff
@@ -25,7 +25,7 @@ import:
   - endpoint
   - transport/http
 - package: github.com/hashicorp/vault
-  version: 0.6.4
+  version: ~0.7.3
```
rossf7: Note the semver

client/vault/vault.go
```diff
@@ -37,7 +36,6 @@ func NewClient(config Config) (*vaultapi.Client, error) {
 
 	newClientConfig := vaultapi.DefaultConfig()
 	newClientConfig.Address = address
-	newClientConfig.HttpClient = http.DefaultClient
```
rossf7: Calling `vaultapi.DefaultConfig()` already configures a HTTP client with transport support. So setting the HTTP client is not necessary and caused an error because `http.DefaultClient` doesn't configure transport support.

https://github.com/hashicorp/vault/blob/master/api/client.go#L89

rossf7:
certctl upgrade has been merged so this is now ready.

PR #63, opened 2017-07-12
Add healthz endpoint that checks the vault token

rossf7:
Fixes giantswarm/giantswarm#1558

Adds healthz endpoint that checks the Vault token and API connection are working. To do this it lists the sys mounts. A health endpoint has been added to the Vault API and will be in the next release. We should switch to that once we've upgraded to Vault 0.7.

https://github.com/hashicorp/vault/blob/master/CHANGELOG.md#073-june-7th-2017


rossf7:
I had problems rebasing the old PR so I did a fresh one. I think this was due to the big vendor update.

@xh3b4sd @kopiczko your comments from #59 are addressed

@JosephSalisbury adding you for info as I added Prometheus support to the healthz endpoint

service/healthz/service.go
```diff
@@ -0,0 +1,102 @@
...
+	micrologger "github.com/giantswarm/microkit/logger"
+	vaultapi "github.com/hashicorp/vault/api"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+var (
+	healthCheckRequests = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "health_check_request_total",
+			Help: "Number of health check requests",
```
xh3b4sd: The help messages are usually sentences. Please add a full stop.

service/healthz/service.go
```diff
@@ -0,0 +1,102 @@
...
+	healthCheckRequestTime = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "health_check_request_milliseconds",
+		Help: "Time taken to respond to health check, in milliseconds",
+	})
+)
+
+func init() {
+	prometheus.MustRegister(healthCheckRequests)
+	prometheus.MustRegister(healthCheckRequestTime)
+}
```
xh3b4sd: I would like to put this into a separate file `metrics.go`.

service/healthz/service.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil, microerror.MaskAnyf(invalidConfigError, "logger must not be empty")
+	}
+	if config.VaultClient == nil {
+		return nil, microerror.MaskAnyf(invalidConfigError, "vault client must not be empty")
+	}
+
+	newService := &Service{
+		// Dependencies.
+		vaultClient: config.VaultClient,
+		logger:      config.Logger,
```
xh3b4sd: Minor nitpick. Alphabetical order.

service/healthz/service.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Check implements the health check which lists the mounts for the sys backend.
+// This checks that the operator can connect to the Vault API and the token is
+// valid.
+func (s *Service) Check(ctx context.Context, request Request) (*Response, error) {
+	start := time.Now()
+	defer func() {
+		healthCheckRequestTime.Set(float64(time.Since(start) / time.Millisecond))
+	}()
```
xh3b4sd: We can do `}(time.Now())`.

service/healthz/service.go
```diff
@@ -0,0 +1,102 @@
...
+
+	sysBackend := s.vaultClient.Sys()
+
+	_, err := sysBackend.ListMounts()
+	if err != nil {
+		healthCheckRequests.WithLabelValues("failed").Inc()
+		return nil, microerror.MaskAny(err)
+	}
+
+	healthCheckRequests.WithLabelValues("successfull").Inc()
```
xh3b4sd: I would like to have constants for this with some godoc describing what it is for. The label is also called `success`. When the actual value for `success` is `failed` it reads weirdly. Maybe the label should be `result`. Uncertain. /cc @JosephSalisbury

service/healthz/service.go
rossf7: Fixed

service/healthz/service.go
rossf7: Done

service/healthz/service.go
rossf7: Fixed, damnit missed that one ;)

service/healthz/service.go
rossf7: I left this as to me it's more readable.

service/healthz/service.go
rossf7: I got rid of the magic strings and added godoc. But to me the label names make sense. As only one is applied and they can be used for filtering.

Although I managed to spell successful wrong. Fail! :D

r7vme:
Did we check this?

Because with latest Vault bootstrap manifests i'm getting

```
curl -H "X-Vault-Token: G8S_VAULT_TOKEN" http://172.16.238.2:8200/v1/sys/mounts
{"errors":["permission denied"]}
```
```
107     -policy="pki-issue-policy-${CLUSTER_NAME}" \
108     -policy="secret-g8s-sa-sign-key" \
109     -policy="default" \
```

xh3b4sd:
@r7vme `permission denied` errors in Vault usually mean the Vault token being used is not valid and probably expired.

PR #70, opened 2017-08-04
updated vendor deps

xh3b4sd:


service/healthz/healthz.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	return newService, nil
+}
+
+// Service is the healthz service collection.
+type Service struct {
+	K8s   healthz.Service
+	Vault healthz.Service
+}
```
xh3b4sd: We have the common endpoints in the `microendpoint` package but I want to have a service collection anyway. So we have to add it. All health checks are initialized here. We can then use it like this. 

```diff
+		healthzConfig.Services = []healthzservice.Service{
+			config.Service.Healthz.K8s,
+			config.Service.Healthz.Vault,
+		}
```

service/healthz/service.go
```diff
@@ -1,73 +0,0 @@
...
-func (s *Service) Check(ctx context.Context, request Request) (*Response, error) {
-	sysBackend := s.vaultClient.Sys()
-
-	_, err := sysBackend.ListMounts()
-	if err != nil {
-		return nil, microerror.MaskAny(err)
-	}
-
-	return DefaultResponse(), nil
-}
```
xh3b4sd: This is gone in favour of the common one.

client/vault/error.go
```diff
@@ -1,12 +1,12 @@
 package vault
 
 import (
-	"github.com/juju/errgo"
+	"github.com/giantswarm/microerror"
```
xh3b4sd: We do the microerror transition here.

main.go
```diff
@@ -4,8 +4,8 @@ import (
 	"os"
 
 	"github.com/giantswarm/microkit/command"
-	"github.com/giantswarm/microkit/logger"
 	microserver "github.com/giantswarm/microkit/server"
+	"github.com/giantswarm/micrologger"
```
xh3b4sd: We also do the micrologger transition here.

PR #77, opened 2017-09-05
Set allowed domains when issuing certs

rossf7:


rossf7:
Closing. Fix doesn't work because some certs don't have alt names.

PR #78, opened 2017-09-05
Set allowed domains from certificate TPO

rossf7:


rossf7:
This fix doesn't work for existing clusters. See #79 for the current fix.

rossf7:
Closing as fix didn't work.

PR #79, opened 2017-09-06
Set allowed domains when issuing certs

rossf7:
This fix adds the allowed domains when issuing certs so the vault role is correct.

### DIFF
--- a/service/crt/cert_signer.go
+++ b/service/crt/cert_signer.go
@@ -1,6 +1,7 @@
 package crt
 
 import (
+	"fmt"
 	"strings"
 
 	"time"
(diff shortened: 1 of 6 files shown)


service/crt/cert_signer.go
```diff
@@ -48,13 +49,21 @@ func (s *Service) Issue(cert certificatetpr.Spec) error {
 		return microerror.Mask(err)
 	}
 
+	allowedDomainsFormat := s.Viper.GetString(s.Flag.Service.Vault.Config.Certificate.AllowedDomainsFormat)
+	if allowedDomainsFormat == "" {
+		return microerror.Maskf(invalidConfigError, "certificate allowed domains must not be empty")
+	}
+	allowedDomains := fmt.Sprintf(allowedDomainsFormat, cert.ClusterID)
```
rossf7: Here we generated the allowed domains.

service/crt/cert_signer.go
```diff
@@ -48,13 +49,21 @@ func (s *Service) Issue(cert certificatetpr.Spec) error {
...
-		IPSANs:     strings.Join(cert.IPSANs, ","),
-		AltNames:   strings.Join(cert.AltNames, ","),
-		TTL:        cert.TTL,
+		ClusterID:        cert.ClusterID,
+		CommonName:       cert.CommonName,
+		IPSANs:           strings.Join(cert.IPSANs, ","),
+		AltNames:         strings.Join(cert.AltNames, ","),
+		TTL:              cert.TTL,
+		AllowedDomains:   allowedDomains,
+		AllowBareDomains: true,
```
rossf7: We also need to set allow bare domains to true for all certs so the Vault role is correct.

PR #82, opened 2017-09-07
add scripts to automate the setup and teardown of the local example

fgimenez:
The setup script is specific for minikube, maybe this could be configured.

### DIFF
--- a/examples/local/env.sh
+++ b/examples/local/env.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-export CLUSTER_NAME=g8s
-export COMMON_DOMAIN=local
-export VAULT_HOST=vault
-export VAULT_TOKEN=mytoken
+export CLUSTER_NAME=${CLUSTER_NAME:-g8s}
+export COMMON_DOMAIN=${COMMON_DOMAIN:-local}
+export VAULT_HOST=${VAULT_HOST:-vault}
+export VAULT_TOKEN=${VAULT_TOKEN:-mytoken}
--- a/examples/local/env.sh
+++ b/examples/local/env.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
 export CLUSTER_NAME=g8s
-export COMMON_DOMAIN=${CLUSTER_NAME}.local
+export COMMON_DOMAIN=local
 export VAULT_HOST=vault
 export VAULT_TOKEN=mytoken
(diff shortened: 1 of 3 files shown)


examples/local/env.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+export CLUSTER_NAME=g8s
+export COMMON_DOMAIN=${CLUSTER_NAME}.local
```
MarcelMue: I don't think this is always the case, `CLUSTER_NAME` and `COMMON_DOMAIN` are completely unrelated in the setup with `aws-operator` I believe.

examples/local/setup-minikube.sh
```diff
@@ -0,0 +1,28 @@
...
+        -e 's|${CLUSTER_NAME}|'"${CLUSTER_NAME}"'|g' \
+        -e 's|${COMMON_DOMAIN}|'"${COMMON_DOMAIN}"'|g' \
+        -e 's|${VAULT_HOST}|'"${VAULT_HOST}"'|g' \
+        -e 's|${VAULT_TOKEN}|'"${VAULT_TOKEN}"'|g' \
+        ./$f > ./${f%.tmpl.yaml}.yaml
+done
+
+kubectl apply -f ./vault.yaml
+eval $(minikube docker-env)
+(
```
MarcelMue: I like this 👍

examples/local/env.sh
rossf7: There is a connection between `cert-operator` and `aws-operator`. The `COMMON_DOMAIN` used should match the one in `aws-operator`. This is so the certs are valid and match the Route 53 domains. 

It also means you can configure `kubectl` to connect to the cluster. I'll update the wiki page to also include `cert-operator` as I missed this step.

examples/local/setup-minikube.sh
rossf7: Same here! 👍

examples/local/env.sh
fgimenez: Ok thank you both, I've updated `COMMON_DOMAIN` to not depend on `CLUSTER_NAME`, let me know if anything else is needed here

examples/local/setup-minikube.sh
fgimenez: Thanks! :)

fgimenez:
I've added two additional changes for building the binary from the local tree and taking into account existing env vars, pls take a look

PR #87, opened 2017-09-13
added k8s settings to configmap

xh3b4sd:


xh3b4sd:
Going with this to fix the broken host clusters. Tested on iris.

PR #91, opened 2017-10-13
migrate PKI backend related resources

xh3b4sd:


glide.yaml
```diff
@@ -14,5 +14,11 @@ import:
   version: master
 - package: github.com/giantswarm/vaulthealthz
   version: master
+- package: github.com/giantswarm/vaultcrt
+  version: init
+- package: github.com/giantswarm/vaultpki
+  version: init
+- package: github.com/giantswarm/vaultrole
+  version: init
```
xh3b4sd: These are the new dependencies. They are leaner and cleaner now. See also 

- https://github.com/giantswarm/vaultcrt/pull/1/files
- https://github.com/giantswarm/vaultpki/pull/1/files
- https://github.com/giantswarm/vaultrole/pull/1/files

service/ca/pki.go
```diff
@@ -1,137 +0,0 @@
-package ca
```
xh3b4sd: I tried to have smaller PRs for this but because of the restructuring it was not really possible to refactor the business logic separately. Thus we remove the old ca and crt magic here and add a `vaultpki` and `vaultcrt` resource to do the magic instead.

service/resource/legacy/error.go
```diff
@@ -1,24 +0,0 @@
-package legacy
```
xh3b4sd: The legacy resource is also removed because the two new resources do the do now.

service/resource/vaultcrt/create.go
```diff
@@ -0,0 +1,123 @@
...
+
+		ca, crt, key, err := r.issueCertificate(customObject)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		secretToCreate.StringData[certificatetpr.CA.String()] = ca
+		secretToCreate.StringData[certificatetpr.Crt.String()] = crt
+		secretToCreate.StringData[certificatetpr.Key.String()] = key
+	}
```
xh3b4sd: Note that we create a role per key-pair now with potential support for organizations. We just do not use this support now because the organizations information is not yet given with the custom object.

service/resource/vaultpki/create.go
```diff
@@ -0,0 +1,70 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "finding out if the Vault PKI has to be created")
+
+	var vaultPKIStateToCreate VaultPKIState
+	if !currentVaultPKIState.BackendExists || !currentVaultPKIState.CAExists {
+		vaultPKIStateToCreate = desiredVaultPKIState
+	}
```
xh3b4sd: As a first step we just work with a custom structure to reflect boolean states for the PKI backend and the root CA. We are still only interested in this binary states because we just want to make sure the PKI backend is setup or not. Later we can potentially extend this easily by tuning the PKI backend mount in Vault. This is out of scope of this PR though.

service/resource/vaultpki/delete.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	desiredVaultPKIState, err := toVaultPKIState(desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	var vaultPKIStateToDelete VaultPKIState
+	if deletionallowedcontext.IsDeletionAllowed(ctx) {
```
xh3b4sd: Note that we do not allow deletion of PKI backends by the operator yet because this is still done by the cluster-service. Reason for this is our insufficient TPR. We will iterate on this in the near future. For now deletion is not allowed because nobody is setting the state to the context and the default is false.

xh3b4sd:
Successfully tested on lycan.

service/resource/vaultcrt/create.go
JosephSalisbury: I don't follow here - what does this have to do with roles?

service/resource/vaultcrt/desired.go
```diff
@@ -0,0 +1,43 @@
...
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "computing the desired secret")
+
+	// NOTE that the actual secret content here is left blank because only the
+	// issuer backend, e.g. Vault, can generate certificates. This has to be
+	// considered when computing the create, delete and update state.
```
JosephSalisbury: So, desired state is the empty string, and we add in the certificates when we actually create them?

service/resource/vaultpki/create.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if vaultPKIStateToCreate.CACertificate != "" {
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "creating the root CA in the Vault PKI")
+
+		err := r.vaultPKI.CreateCA(key.ClusterID(customObject))
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "created the root CA in the Vault PKI")
```
JosephSalisbury: this reads oddly to me - it's the vault pki backend, right?

service/resource/vaultpki/create_test.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+func Test_Resource_VaultPKI_GetCreateState(t *testing.T) {
+	testCases := []struct {
+		Obj           interface{}
+		CurrentState  interface{}
+		DesiredState  interface{}
+		ExpectedState VaultPKIState
+	}{
+		{
```
JosephSalisbury: Some comments on these test cases would be appreciated

service/resource/vaultpki/desired_test.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/giantswarm/vaultpki/vaultpkitest"
+	vaultapi "github.com/hashicorp/vault/api"
+)
+
+func Test_Resource_VaultPKI_GetDesiredState(t *testing.T) {
+	testCases := []struct {
+		Obj           interface{}
+		ExpectedState VaultPKIState
+	}{
+		{
```
JosephSalisbury: I'd appreciate some test descriptions here

service/resource/vaultpki/resource.go
```diff
@@ -0,0 +1,68 @@
...
+func New(config Config) (*Resource, error) {
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty")
+	}
+	if config.VaultPKI == nil {
+		return nil, microerror.Maskf(invalidConfigError, "config.VaultPKI must not be empty")
+	}
+
+	r := &Resource{
+		logger: config.Logger.With(
```
JosephSalisbury: This is a nice touch

service/resource/vaultcrt/create.go
```diff
@@ -0,0 +1,127 @@
...
+	desiredSecret, err := toSecret(desiredState)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "finding out if the secret has to be created")
+
+	var secretToCreate *apiv1.Secret
+	if currentSecret == nil {
```
kopiczko: Nitpick: I almost always prefer short `if` branches. There is less information to carry when looking at the code. In this case it would be

```
if currentSecret != nil {
        r.logger.Log("cluster", key.ClusterID(customObject), "debug", "found out secret is already created")
        return nil, nil
}
```

service/resource/vaultcrt/delete.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "finding out if the secret has to be deleted")
+
+	var secretToDelete *apiv1.Secret
+	if currentSecret != nil {
+		secretToDelete = currentSecret
+	}
+
+	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "found out if the secret has to be deleted")
```
kopiczko: I think we can reduce one log by just saying that we found out secret has to be be deleted or is deleted already.

service/resource/vaultpki/desired.go
```diff
@@ -0,0 +1,37 @@
...
+	// reconciliation. It is also that in case of the CA certificate we cannot
+	// just predict and define it here, because this is the resonsibility of the
+	// actual issuer backend, e.g. Vault.
+	var vaultPKIState VaultPKIState
+	{
+		vaultPKIState.Backend = &vaultapi.MountOutput{
+			Type: "pki",
+		}
+
+		vaultPKIState.CACertificate = "placeholder"
```
kopiczko: How is this placeholder replaced?

service/resource/vaultcrt/create.go
xh3b4sd: I agree with short branches but our definition of short seems to be different. The branch is 15 lines long. This is easily observable on almost every "normal" screen you work on. So this is still ok for me. Further you mean early returns which I have no strong feelings about at this point.

service/resource/vaultcrt/create.go
xh3b4sd: This is just how Vault works. There has to be a Vault role for each certificate we want to issue via the Vault PKI backend.

service/resource/vaultcrt/delete.go
xh3b4sd: I think we can reduce logging in general. I am personally still in the middle of learning how to implement resources and which information are important to log. For now I would like to keep the scheme I used in all of my implementations and clean them up as soon as I know how it should be. Making a little change here and later there will only cause snowflaky logging and this would make me super sad.

service/resource/vaultcrt/desired.go
xh3b4sd: Yes, reason is that we do not want to generate certificates every five minutes to not use them at the end. This is just due to the fact how certificate generation works in general.

service/resource/vaultpki/create.go
xh3b4sd: Yes. Would it be better to add the word `backend` here?

service/resource/vaultpki/desired.go
xh3b4sd: Never, it is only to have some content as desired state. The CA will be generated in `ProcessCreateState` in case the `CACertificate` property is not empty. This mimics the state management of the usual reconciliation because it is a bit different when working with certificate generation.

service/resource/vaultpki/create_test.go
xh3b4sd: Done.

service/resource/vaultpki/desired_test.go
xh3b4sd: Done.

xh3b4sd:
Please have another look guys. Fishing for LGTM.

service/resource/vaultpki/create.go
JosephSalisbury: Not fussed. Do it if you care, I get it now :D

service/resource/vaultpki/current.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	var vaultPKIState VaultPKIState
+
+	{
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "looking for the Vault PKI in the Vault API")
+
+		vaultPKIState.Backend, err = r.vaultPKI.GetBackend(key.ClusterID(customObject))
+		if vaultpki.IsNotFound(err) {
+			r.logger.Log("cluster", key.ClusterID(customObject), "debug", "did not find the Vault PKI in the Vault API")
```
kopiczko: We shouldn't rely on the fact that `GetBackend` returns `nil` backend when error is not `nil`

service/resource/vaultpki/current.go
```diff
@@ -0,0 +1,47 @@
...
+			r.logger.Log("cluster", key.ClusterID(customObject), "debug", "found the Vault PKI in the Vault API")
+		}
+	}
+
+	{
+		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "looking for the root CA in the Vault PKI")
+
+		vaultPKIState.CACertificate, err = r.vaultPKI.GetCACertificate(key.ClusterID(customObject))
+		if vaultpki.IsNotFound(err) {
+			r.logger.Log("cluster", key.ClusterID(customObject), "debug", "did not find the root CA in the Vault PKI")
```
kopiczko: We shouldn't rely on the fact that GetCACertificate returns nil CACertificate when error is not nil

service/resource/vaultpki/spec.go
```diff
@@ -0,0 +1,8 @@
+package vaultpki
+
+import vaultapi "github.com/hashicorp/vault/api"
+
+type VaultPKIState struct {
+	Backend       *vaultapi.MountOutput
+	CACertificate string
```
kopiczko: It took me a lot of time to wrap my head around what is going on here. This `string` serves a `bool` function in this line:

```
	if vaultPKIStateToCreate.CACertificate != "" {
```

The `GetDesiredState` function doesn't work in this resource, and let's not try to hide that. I suggest changing this field to `CACertificateExists bool`. That way we can get rid of placeholders and it'll become obvious when the cert is actually created, and what is the function of this field.

xh3b4sd: > This `string` serves a bool function in this line. 

No it does not. We need three states and not two. I started working with `BackendExists` and `CAExists` and ran into problems that either create, delete or update screws the reconciliation. The default would be `false` on which we then created them during `ProcessCreateState`. There is no update procedure implemented so the zero value is returned by `GetUpdateState`. `ProcessCreateState` got called again and saw `false` and tried to create the resources again, but they already existed and the reconciliation failed with "already exists" errors. Having explicit state which we actively set to something enables us to reflect the requirements for creation, deletion and updates. I know this sucks but at this point I do not know how to improve this. Some resources are apparently just different and we need to deal with their specialties somehow.

service/resource/vaultpki/current.go
xh3b4sd: Done.

service/resource/vaultpki/current.go
xh3b4sd: It returns an empty string, not nil, but I get you. Done.

service/resource/vaultpki/spec.go
kopiczko: I miss something in the flow. This is how I understand it.

First loop:

1. GetCreateState returns `CAExists=false`.
2. ProcessCreateState creates CA.
3. GetUpdateState returns zero object.
4. ProcessUpdate does nothing.
5. End of reconciliation loop.

Next loop:

1. GetCreateState returns `CAExists=true`.
2. ProcessCreateState does nothing.
3. GetUpdateState returns zero object.
4. ProcessUpdate does nothing.
5. End of reconciliation loop.

I don't understand how `ProcessCreateState` can be called with state created by `GetUpdateState`.

xh3b4sd: It is this. 

First loop:

1. GetCreateState returns CAExists=false.
2. ProcessCreateState creates CA.
3. GetUpdateState returns zero object (CAExists=false).
4. ProcessUpdateState does nothing.
5. ProcessCreateState creates CA and fails with already exists.

xh3b4sd: See https://github.com/giantswarm/operatorkit/blob/57dd6cec0cc811e2bc2e2c8a56ea40f163673514/framework/framework.go#L342-L398.

kopiczko: I see. IMO this means `GetUpdateState` should be return correct `createState`.

xh3b4sd: I don't like this.

kopiczko: Relying on zero value of:

```
	if vaultPKIStateToCreate.CACertificate != "" {
```

is like relying on zero value of

```
	if vaultPKIStateToCreate.CACertificateNotExists {
```

which I don't like.

In that case zero value of `CACertificateNotExists` is `false` which means CA certificate exists.  I still think we use `string` as `bool` and `GetUpdateState` implementation is incorrect. As long as `ProcessUpdate` invokes `PorcessCreate` inside the framework, `GetUpdateState` should always return computed `createState` IMO.

xh3b4sd: It is not boolean. We have three states because deletion handles state reversed as opposed to creation. The framework is not perfect but I also do not know how to improve it right now. There is no simple answer to this because there are a lot of complex cases that the framework should consider. I want to implement creation and deletion in case the resource requires creation and deletion based on its nature. Being forced to implement updates even though the resource does not require it only to make the framework work is just wrong to me. I also thought about resources that you provide only the creation and/or deletion interface to not even need to implement the empty update methods. This would then not work in this case, but it should. 

I like your efforts on doing the right thing but I do not want to get lost in details now. Feel free to propose changes to the framework to make the implementations easier. For now I would like to go with what we have to solve real problems and make our customers happy. This refactoring here eats too much time already. Lets iterate and do the right thing as soon as we know what the right thing is.

xh3b4sd:
I am going with this now as a first step. The library was used by the refactored `cert-operator` for over a week now on `gauss`, `ginger` and `lycan`.

Issue #95, opened 2017-10-23
Allow certificate creation in namespaces


r7vme:
Currently all secrets created in default namespace. I want to be able to control this behaviour.

Current behaviour. I created TPO in namespace, but secrets created in default namespace.
```
apiVersion: giantswarm.io/v1
kind: Certificate
metadata:
  labels:
    clusterComponent: node-controller
    clusterID: efk16
  name: efk16-node-controller
  namespace: efk16
spec:
  allowBareDomains: true
  clusterComponent: node-controller
  clusterID: efk16
  commonName: node-controller.efk16.g8s.fra-1.giantswarm.io
  ttl: 4320h
```
```
~|⇒ kubectl get certificate -n efk16                          
NAME                    KIND
efk16-node-controller   Certificate.v1.giantswarm.io
~|⇒ kubectl get secrets -n default | grep efk16-node-controller
efk16-node-controller   Opaque                                3         11s
```

Expected behaviour:
If `spec ->namespace` is set create resulted secrets in specified namespace.

xh3b4sd:
Feel free to go ahead on your own. I guess you need to add the issuance for your new cert/secret anyway. See https://github.com/giantswarm/kubernetesd/blob/d2de4b238885dbb05f75046000bf57b0c47d8272/service/creator/certificate.go. 

Would be good to add the namespace property to https://github.com/giantswarm/certificatetpr. 

The operator can then create the secret in the right namespace according to the given custom object. See https://github.com/giantswarm/cert-operator/blob/e7c7f2923ddc2fd3cfb5c76ade467a2e40b208e4/service/resource/vaultcrt/create.go#L68.

r7vme:
We decided to not implement this.

PR #96, opened 2017-10-24
Add namespace handling

r7vme:
Tests will fail due to not vendored certificatetpr. https://github.com/giantswarm/certificatetpr/pull/29

If secret namespace set in TPO then use one
if not then use defalt.

Partly implements: https://github.com/giantswarm/cert-operator/issues/95
Depends on: https://github.com/giantswarm/certificatetpr/pull/29

### DIFF
--- a/service/resource/vaultcrt/current.go
+++ b/service/resource/vaultcrt/current.go
@@ -21,7 +21,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	var secret *apiv1.Secret
 	{
-		manifest, err := r.k8sClient.Core().Secrets(r.namespace).Get(key.SecretName(customObject), apismetav1.GetOptions{})
+		ns := key.SecretNamespace(customObject)
+		if ns == "" {
+			ns = r.namespace
+		}
+
+		manifest, err := r.k8sClient.Core().Secrets(ns).Get(key.SecretName(customObject), apismetav1.GetOptions{})
 		if apierrors.IsNotFound(err) {
 			r.logger.Log("cluster", key.ClusterID(customObject), "debug", "did not find the secret in the Kubernetes API")
 			// fall through
--- a/service/resource/vaultcrt/delete.go
+++ b/service/resource/vaultcrt/delete.go
@@ -46,7 +46,12 @@ func (r *Resource) ProcessDeleteState(ctx context.Context, obj, deleteState inte
 	if secretToDelete != nil {
 		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "deleting the sercet in the Kubernetes API")
 
-		err = r.k8sClient.CoreV1().Secrets(r.namespace).Delete(secretToDelete.Name, &apismetav1.DeleteOptions{})
+		ns := key.SecretNamespace(customObject)
+		if ns == "" {
+			ns = r.namespace
+		}
+
+		err = r.k8sClient.CoreV1().Secrets(ns).Delete(secretToDelete.Name, &apismetav1.DeleteOptions{})
 		if apierrors.IsNotFound(err) {
 			// fall through
 		} else if err != nil {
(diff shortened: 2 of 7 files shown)


service/resource/vaultcrt/current.go
```diff
@@ -19,9 +19,14 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 
 	r.logger.Log("cluster", key.ClusterID(customObject), "debug", "looking for the secret in the Kubernetes API")
 
+	ns := key.SecretNamespace(customObject)
+	if ns == "" {
+		ns = r.namespace
+	}
```
xh3b4sd: Since this is only used within the scope block below I would like to put this code inside the block.

service/resource/vaultcrt/delete.go
```diff
@@ -46,6 +46,11 @@ func (r *Resource) ProcessDeleteState(ctx context.Context, obj, deleteState inte
 	if secretToDelete != nil {
 		r.logger.Log("cluster", key.ClusterID(customObject), "debug", "deleting the sercet in the Kubernetes API")
 
+		ns := key.SecretNamespace(customObject)
+		if ns == "" {
+			ns = r.namespace
+		}
+
 		err = r.k8sClient.CoreV1().Secrets(r.namespace).Delete(secretToDelete.Name, &apismetav1.DeleteOptions{})
```
xh3b4sd: Lets make sure we use the right namespace for deletion.

r7vme: Crap, yes.

service/resource/vaultcrt/current.go
r7vme: yes.

r7vme:
@kopiczko need more discussion i think.

Before we did not assume that TPO will affect namespace. Now we have the case where we want to have resource (secret) in namespace. And for me solution that we have now (setting namespace in TPO) is more like hack.

but do we really need universal flexible namespace management. e.g. TPO namespace --> Resource namespace. or smth like that.

Other thing, i can not build cert-operator after `glide up -v` with . Do you have idea what can be wrong? 
```
cert-operator|add-namespace-handle ⇒ go test ./...
# github.com/giantswarm/cert-operator/service/resource/vaultpki
../../src/go/src/github.com/giantswarm/cert-operator/service/resource/vaultpki/resource.go:54:2: cannot use r (type *Resource) as type framework.Resource in return argument:
        *Resource does not implement framework.Resource (missing ApplyCreateChange method)
# github.com/giantswarm/cert-operator/service/resource/vaultcrt
../../src/go/src/github.com/giantswarm/cert-operator/service/resource/vaultcrt/resource.go:86:2: cannot use r (type *Resource) as type framework.Resource in return argument:
        *Resource does not implement framework.Resource (missing ApplyCreateChange method)
# github.com/giantswarm/cert-operator/service/resource/vaultpki
../../src/go/src/github.com/giantswarm/cert-operator/service/resource/vaultpki/resource.go:54:2: cannot use r (type *Resource) as type framework.Resource in return argument:
        *Resource does not implement framework.Resource (missing ApplyCreateChange method)
# github.com/giantswarm/cert-operator/service/resource/vaultcrt
../../src/go/src/github.com/giantswarm/cert-operator/service/resource/vaultcrt/resource.go:86:2: cannot use r (type *Resource) as type framework.Resource in return argument:
        *Resource does not implement framework.Resource (missing ApplyCreateChange method
```

xh3b4sd:
@r7vme I would say you pin the former operatorkit version since we just merged quite a big change and breaking change. You fetched the breaking change and just experienced how important it is to version our code. :D

r7vme:
yep, will do version pinning. 

So far are we ok with overwriting "global" resource namespace this way(like in this PR)? @xh3b4sd @kopiczko

xh3b4sd:
I think it is good to have all TPOs of a cluster in the same namespace. So this "hack" works for me.

r7vme:
It was decided to not implement this.

PR #98, opened 2017-10-26
vendor operatorkit changes

kopiczko:
Towards https://github.com/giantswarm/giantswarm/issues/2056


service/resource/vaultcrt/create.go
```diff
@@ -12,7 +12,33 @@ import (
 	"github.com/giantswarm/cert-operator/service/key"
 )
 
-func (r *Resource) GetCreateState(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error) {
+func (r *Resource) ApplyCreateChange(ctx context.Context, obj, createChange interface{}) error {
```
kopiczko: * renamed `Process*State` -> `Apply*Change` and moved up
* renamed `Get*State` -> `new*Change` and moved down

service/resource/vaultcrt/delete.go
```diff
@@ -11,34 +11,12 @@ import (
 	"github.com/giantswarm/cert-operator/service/key"
 )
```
kopiczko: * renamed `Process*State` -> `Apply*Change` and moved up
* renamed `Get*State` -> `new*Change` and moved down

service/resource/vaultpki/create.go
```diff
@@ -8,41 +8,12 @@ import (
 	"github.com/giantswarm/cert-operator/service/key"
 )
```
kopiczko: * renamed `Process*State` -> `Apply*Change` and moved up
* renamed `Get*State` -> `new*Change` and moved down

service/resource/vaultpki/delete.go
```diff
@@ -9,7 +9,33 @@ import (
 	"github.com/giantswarm/cert-operator/service/key"
 )
```
kopiczko: * renamed `Process*State` -> `Apply*Change` and moved up
* renamed `Get*State` -> `new*Change` and moved down

service/resource/vaultcrt/patch.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	patch := framework.NewPatch()
+	patch.SetCreateChange(create)
+	patch.SetUpdateChange(update)
+
+	return patch, nil
+}
+
+func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desiredState interface{}) (*framework.Patch, error) {
```
xh3b4sd: I would like to keep these methods in the update and delete files. Adding a new patch file does not feel like the right thing to me.

kopiczko: They aren't in 1:1 relation. Any `New*Patch` may use any `new*Change`. I wouldn't feel comfortable with that.

xh3b4sd: It is not about where they are used but what they do. The creation magic belongs to the create file. The update magic belongs to the update file. The deletion magic belongs to the delete file. It was like this and I want to keep it like this.

kopiczko: Where NewUpdatePatch belongs to when it references create and update magic?

xh3b4sd: The result is the update patch. So it belongs to the update file IMO.

kopiczko: I think those are separate things. `create changes` may be different for `update patch` and `delete patch`. Both update & delete patches may contain all types of changes. To name a few cases. I treat `patch.go` as a dispatcher of changes which has to be applied. I can disagree and commit though.

xh3b4sd: When they are different there should be something like `newCreateChangeForDeletePatch`. This should then be within the create file, because it is about the create state. 

> I don't like asymmetry in create.go file.

For the the asymmetry is there with the patch file. It is just a new file that makes everything different and I cannot see the clear line anymore. I think the rule is using the file defined by `*` when having the scheme `New*Patch` or `Apply*Change`. That is a pretty good guide to decide where to put code.

kopiczko: I don't like it. But I disagree and commit to make this going.
Done.

PR #103, opened 2017-11-09
Add organization

jgsqware:
I've updated the certificatetrp deps

https://github.com/giantswarm/giantswarm/issues/2062


service/resource/vaultcrt/create.go
```diff
@@ -96,9 +96,8 @@ func (r *Resource) ensureVaultRole(customObject certificatetpr.CustomObject) err
 			AllowSubdomains:  AllowSubDomains,
 			AltNames:         key.AltNames(customObject),
 			ID:               key.ClusterID(customObject),
-			Organizations: []string{
-				key.ClusterComponent(customObject),
-			},
+			Organizations: append([]string{key.ClusterComponent(customObject)},
+				key.Organizations(customObject)...),
```
xh3b4sd: Above has to be done the same thing. It is not nice. There is an issue somewhere to make this more convenient.

service/key/key.go
```diff
@@ -35,6 +35,10 @@ func IPSANs(customObject certificatetpr.CustomObject) []string {
 	return customObject.Spec.IPSANs
 }
 
+func Organizations(customObject certificatetpr.CustomObject) []string {
+	return customObject.Spec.Organizations
+}
```
xh3b4sd: Lets please do the append magic do in here. The organizations needed are the cluster component and the configured orgs in the tpo.

jgsqware:
@xh3b4sd Is this better?

service/resource/vaultcrt/create.go
```diff
@@ -111,15 +109,14 @@ func (r *Resource) ensureVaultRole(customObject certificatetpr.CustomObject) err
 }
 
 func (r *Resource) issueCertificate(customObject certificatetpr.CustomObject) (string, string, string, error) {
+
```
xh3b4sd: Why that newline?

service/resource/vaultcrt/create.go
```diff
@@ -96,10 +96,8 @@ func (r *Resource) ensureVaultRole(customObject certificatetpr.CustomObject) err
 			AllowSubdomains:  AllowSubDomains,
 			AltNames:         key.AltNames(customObject),
 			ID:               key.ClusterID(customObject),
-			Organizations: []string{
-				key.ClusterComponent(customObject),
-			},
-			TTL: key.RoleTTL(customObject),
+			Organizations:    key.Organizations(customObject),
```
xh3b4sd: I like this ease, but as I said before, we also have to set the orgs above when checking if the role exists.

jgsqware: Good catch haven't seen it :)

PR #105, opened 2017-11-13
retry framework forever

xh3b4sd:
See https://github.com/giantswarm/kvm-operator/pull/243.

### DIFF
--- a/service/service.go
+++ b/service/service.go
@@ -244,7 +244,7 @@ func New(config Config) (*Service, error) {
 	var frameworkBackOff *backoff.ExponentialBackOff
 	{
 		frameworkBackOff = backoff.NewExponentialBackOff()
-		frameworkBackOff.MaxElapsedTime = 5 * time.Minute
+		frameworkBackOff.MaxElapsedTime = 0 // retry forever
 	}
 
 	var operatorFramework *framework.Framework


service/service.go
```diff
@@ -244,7 +244,7 @@ func New(config Config) (*Service, error) {
 	var frameworkBackOff *backoff.ExponentialBackOff
 	{
 		frameworkBackOff = backoff.NewExponentialBackOff()
-		frameworkBackOff.MaxElapsedTime = 5 * time.Minute
+		frameworkBackOff.MaxElapsedTime = 0 // retry forever
```
kopiczko: I think it'd be nice to use `WithMaxTries`, so the operator fails at some point when something is wrong and we are alerted.

xh3b4sd: Lets say we use 10 max tries. The operator runs and every now and then an error occurs because of some network issues. This is not that serious and the operator can continue retrying. Using max retries such situation can happen 10 times and then the operator fails even though it could still proceed just because the max tries are reached. I am not certain there is any reset mechanism in the backoff implementation nor we should rely on this. I think we should retry forever and get alerted based on error rates of the operators.

kopiczko: I didn't give it enough read. I thought it's used to create the framework instance, but I recon it is used to process events.

Now I'm uncertain if we should reuse the same backoff for more than a single operation. I think the library is designed like one backoff for one operation.
Why not using backoff factory function here?

xh3b4sd: That would be good actually. I would like to do this then when we unified all the operator boilerplate. We are currently going through all the operators to fix the same thing, which is super tiring.

xh3b4sd: See https://github.com/giantswarm/giantswarm/issues/2119.

xh3b4sd: Going with this now to fix the problem quick and dirty. Will try to do some work on operatorkit to unify the operator magic for all operators.

kopiczko: Thank you @xh3b4sd  <3 Feel free to merge those changes after the CI passes. They are LGTM.
I can help you with that after Magic planning. Let me know.

PR #106, opened 2017-11-13
lower informer rate wait to speed up reconciliation

xh3b4sd:


xh3b4sd:
Vol saw some issues creating new clusters wile we deployed the cert-operator a lot. On every boot the reconciliation is done again. This seems to block the reconciliation of add events a bit due to locking the add function in the framework, which is not that deterministic for add events. Lowering the rate wait should help a bit relaxing such situations because all loops are done within a way shorter timespan.

PR #107, opened 2017-11-14
Updates vendor

JosephSalisbury:
Towards https://github.com/giantswarm/giantswarm/issues/1796


service/service.go
```diff
@@ -304,6 +289,22 @@ func New(config Config) (*Service, error) {
...
 
+	var operatorFramework *framework.Framework
+	{
+		frameworkConfig := framework.DefaultConfig()
+
+		frameworkConfig.BackOffFactory = framework.DefaultBackOffFactory()
+		frameworkConfig.Informer = newInformer
+		frameworkConfig.InitCtxFunc = initCtxFunc
+		frameworkConfig.Logger = config.Logger
+		frameworkConfig.ResourceRouter = framework.DefaultResourceRouter(resources)
```
xh3b4sd: The TPR should be configured as well to make sure it is created on boot.

JosephSalisbury: Done

PR #108, opened 2017-11-15
use unified operatorkit framework

xh3b4sd:


xh3b4sd:
Moving since it is basically a vendor update.

PR #111, opened 2017-11-16
Fix organization

jgsqware:
Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>

`append()` was messed up due to the `sort` method in the `vaultrole.Exists` method.

Creating a new array fix the problem. Thanks @JosephSalisbury


jgsqware:
@xh3b4sd what do you think for the test?

service/resource/vaultcrt/create_test.go
```diff
@@ -128,3 +129,50 @@ func Test_Resource_VaultCrt_newCreateChange(t *testing.T) {
...
+		resourceConfig.VaultRole = vaultroletest.New()
+
+		newResource, err = New(resourceConfig)
+		if err != nil {
+			t.Fatal("expected", nil, "got", err)
+		}
+	}
+
+	for i, tc := range testCases {
+		err := newResource.ensureVaultRole(tc.CustomObject)
```
xh3b4sd: For my interest, does the test fail when you remove your actual fix from `key.Organizations`? Why do we need to test `VaultCrt.ensureVaultRole` when the fix is in `key.Organizations`? I thought the test could be something like this. 

```
check
key.Organizations(tc.CustomObject)
check
key.Organizations(tc.CustomObject)
check
key.Organizations(tc.CustomObject)
check
```

service/resource/vaultcrt/create_test.go
```diff
@@ -128,3 +129,50 @@ func Test_Resource_VaultCrt_newCreateChange(t *testing.T) {
...
+		{
+			CustomObject: certificatetpr.CustomObject{
+				Spec: certificatetpr.Spec{
+					ClusterID:        "foobar",
+					ClusterComponent: "api",
+					Organizations:    []string{"system:masters"},
+				},
+			},
+			ExpectedOrganizations: []string{"api", "system:masters"},
+		},
```
xh3b4sd: I would like to see multiple tests with different values, empty lists, lists with one and multiple items, also in different orders.

service/resource/vaultcrt/create_test.go
```diff
@@ -128,3 +129,50 @@ func Test_Resource_VaultCrt_newCreateChange(t *testing.T) {
...
+		}
+	}
+
+	for i, tc := range testCases {
+		err := newResource.ensureVaultRole(tc.CustomObject)
+		if err != nil {
+			t.Fatal("case", i, "expected", nil, "got", err)
+		}
+		if !reflect.DeepEqual(tc.ExpectedOrganizations, key.Organizations(tc.CustomObject)) {
+			t.Fatalf("case %d expected %#v got %#v", i, tc.ExpectedOrganizations, key.Organizations(tc.CustomObject))
```
xh3b4sd: This is the function that got the actual fix. I guess the test should be around this method. Also, when you want to test its behaviour because it was wrong before it should be called once and the result should be validated after using it. Calling it twice can still just produce the wrong behaviour without us noticing. So we should call it, set the result to a variable and use the variable for the inspection.

jgsqware: I will try to make a test with the sort method
Will see this tomorrow

jgsqware:
@xh3b4sd WDYT?

service/key/key_test.go
```diff
@@ -0,0 +1,55 @@
...
+		{
+			Organizations:         []string{"system:master"},
+			ClusterComponent:      "api",
+			ExpectedOrganizations: []string{"api", "system:master"},
+		},
+		{
+			Organizations:         []string{"system:master", "giantswarm"},
+			ClusterComponent:      "api",
+			ExpectedOrganizations: []string{"api", "system:master", "giantswarm"},
+		},
```
xh3b4sd: Lets have another test with the same organizations but ordered the other way around.

xh3b4sd: Do the tests actually break when the fix to the fixed function is removed?

jgsqware: No, and I cannot reproduced it at all :(

jgsqware: Sure, But i don't know how to reproduce it. 've created the same case in a dummy project ( with struct, method and so on. and cannot reproduce :/

xh3b4sd: Does that mean when you deploy cert-operator with master everything is fine again?

jgsqware: No, the bug is there. But cannot reproduce it in the test.

jgsqware: I had the sort stuff there. but anyway it's not reproducing the bug we found

jgsqware:
WDYT? could we go with this?

xh3b4sd:
I don't know. Does it solve the problem? You said you cannot reproduce it anymore and I am not sure what the actual state is now. Maybe describe where we are now. I am just confused at this point.

jgsqware:
So, in production, without the fix, the problem persist. So this fix the bug seen yesterday.
However, I cannot reproduce it via Testing.

xh3b4sd:
That it fixes the problem we saw in production is good. We should find out how to test this though to be able to verify the fix programatically.

jgsqware:
So for that we need to reproduce programmatically the problem. And that I'm not able to do it

xh3b4sd:
Then get some help from your team first. Maybe pair on it. You have to escalate that then.

rossf7:
@xh3b4sd @jgsqware I've reproduced the problem locally with cert-operator and this PR definitely fixes the problem! Without the fix

```
openssl x509 -in certs/apiserver.crt -noout -subject
subject= /O=api/CN=api.test-cluster.aws.gigantic.io
```

With the fix

```
openssl x509 -in certs/apiserver.crt -noout -subject
subject= /O=api/O=system:masters/CN=api.test-cluster.aws.gigantic.io
```

From debugging the problem is because without the fix key.Organizations returns "api,api" instead of "api,system:masters". The fix works because it always returns a new string slice instead of appending to the existing slice.

However I can't see a way of updating the test to verify that. Does anyone have an idea on how to test this? As I'm stuck on this now :(

We'd really like to get the fix merged so we can do a full test on `gauss` of `aws-operator` with the `operatorkit` branch that includes the latest k8scloudconfig. CC @fgimenez

fgimenez:
> We'd really like to get the fix merged so we can do a full test on gauss of aws-operator with the operatorkit branch that includes the latest k8scloudconfig.

Definitely, we are not able yet to have a proper e2e test to verify that the fix works (very soon hopefully), not sure if the effects of the fix can be verified at the unit level. If not IMO it would be better to go ahead taking into account the manual verification and have the e2e check in place later to prevent regressions.

rossf7:
Synced with Tim on this and I'm going to update the test so we check both the results of key.Organizations and that Organizations in the custom object doesn't change.

We'll then go with the fix as it's been verified twice now. Once we have e2e tests and can check guest clusters launch correctly it will make this testing less manual and painful!

service/key/key_test.go
```diff
@@ -0,0 +1,60 @@
+package key
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/giantswarm/certificatetpr"
+)
+
+func Test_Organization_sort(t *testing.T) {
```
xh3b4sd: Lets call this `Test_Organizations`.

service/key/key_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	for i, tc := range testCases {
+		customObject := certificatetpr.CustomObject{
+			Spec: certificatetpr.Spec{
+				ClusterComponent: tc.ClusterComponent,
+				Organizations:    tc.Organizations,
+			},
+		}
+
+		result := Organizations(customObject)
```
xh3b4sd: Lets please do this verification and the call to the tested function twice.

fgimenez:
@rossf7 @xh3b4sd @jgsqware Added test that verifies the fix (fails without it), the problem is that somehow the custom object's organizations slice has a capacity of 4 when running on the actual operator, the copy created in `key.Organizations` is sorted here https://github.com/giantswarm/vaultrole/blob/master/key/key.go#L43 (from `ensureCreateRole` in the operator) and this modifies the original organizations slice in the custom object (it still has length 1 but after sorting the underlying array the first element can be different), details in the test https://github.com/giantswarm/cert-operator/pull/111/commits/bf0e930c64389ebe1ce658286c0e6a4d1af5b7d7

jgsqware:
Excellent Federico! Thanks for this deep analysis

PR #112, opened 2017-11-16
renew certs within Kubernetes secrets before they expire

xh3b4sd:


flag/service/resource/vaultcrt/vaultrt.go
```diff
@@ -0,0 +1,6 @@
+package vaultcrt
+
+type VaultCrt struct {
+	ExpirationThreshold string
+	Namespace           string
+}
```
xh3b4sd: This is trying to make it right in the first place, plus boyscouting. ExpirationThreshold is a new option and Namespace already existed but was only hard coded in the resource implementation.

service/resource/vaultcrt/create.go
```diff
@@ -77,48 +75,3 @@ func (r *Resource) newCreateChange(ctx context.Context, obj, currentState, desir
 
 	return secretToCreate, nil
 }
-
-func (r *Resource) ensureVaultRole(customObject certificatetpr.CustomObject) error {
```
xh3b4sd: This is now used for creation and deletion, so I moved it to the `resource.go` to have the code in a more common place.

service/resource/vaultcrt/create_test.go
```diff
@@ -104,14 +105,18 @@ func Test_Resource_VaultCrt_newCreateChange(t *testing.T) {
 	var err error
 	var newResource *Resource
 	{
-		resourceConfig := DefaultConfig()
+		c := DefaultConfig()
```
xh3b4sd: Boyscouting.

service/resource/vaultcrt/create_test.go
```diff
@@ -104,14 +105,18 @@ func Test_Resource_VaultCrt_newCreateChange(t *testing.T) {
...
-		resourceConfig.VaultRole = vaultroletest.New()
+		c.CurrentTimeFactory = func() time.Time { return time.Time{} }
+		c.K8sClient = fake.NewSimpleClientset()
+		c.Logger = microloggertest.New()
+		c.VaultCrt = vaultcrttest.New()
+		c.VaultRole = vaultroletest.New()
 
-		newResource, err = New(resourceConfig)
+		c.ExpirationThreshold = 24 * time.Hour
+		c.Namespace = "default"
```
xh3b4sd: The new flags are required.

service/resource/vaultcrt/resource.go
```diff
@@ -1,55 +1,75 @@
...
 )
 
 type Config struct {
-	K8sClient kubernetes.Interface
-	Logger    micrologger.Logger
-	VaultCrt  vaultcrt.Interface
-	VaultRole vaultrole.Interface
-
-	Namespace string
+	CurrentTimeFactory func() time.Time
```
xh3b4sd: I had to add this to keep the code testable.

service/resource/vaultcrt/update.go
```diff
@@ -25,10 +56,80 @@ func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desire
...
+	if !ok {
+		return false, microerror.Mask(missingAnnotationError)
+	}
+
+	t, err := time.ParseInLocation(UpdateTimestampLayout, a, time.UTC)
+	if err != nil {
+		return false, microerror.Mask(err)
+	}
+
+	if t.Add(TTL).Add(-threshold).Before(r.currentTimeFactory()) {
```
xh3b4sd: `t` is the time of the last update. `t` plus the certificate TTL is the point in time where the certificate expires. We want to renew certificates before they expire. Thus we substitute a threshold from the point in time of the expiration and compare this value against the current time.

helm/cert-operator-chart/templates/configmap.yaml
```diff
@@ -16,6 +16,10 @@ data:
           caFile: ''
           crtFile: ''
           keyFile: ''
+      resource:
+        vaultCrt:
+          expirationThreshold: '24h'
+          namespace: 'default'
```
xh3b4sd: The new settings are properly set in the config map. Note that the 24 hours is a magic number I just made up. IMO it should be good enough to let the cert-operator know 24 hours before the certificates expire that it should actually renew them in the Kubernetes secrets.

service/resource/vaultcrt/resource.go
```diff
@@ -1,55 +1,75 @@
...
 	// AllowSubDomains defines whether to allow the generated root CA of the PKI
 	// backend to allow sub domains as common names.
 	AllowSubDomains = true
+	Name            = "vaultcrt"
+	// UpdateTimestampAnnotation is the annotation key used to track the last
+	// update timestamp of certificates contained in the Kubernetes secrets.
+	UpdateTimestampAnnotation = "giantswarm.io/update-timestamp"
+	// UpdateTimestampLayout is the time layout used to format and parse the
+	// update timestamps tracked in the annotations of the Kubernetes secrets.
+	UpdateTimestampLayout = "2006-01-02T15:04:05.000000Z"
```
JosephSalisbury: is `Layout` the right term here? I was confused here. I think the better term is `Format`

service/resource/vaultcrt/update.go
```diff
@@ -25,10 +56,80 @@ func (r *Resource) NewUpdatePatch(ctx context.Context, obj, currentState, desire
...
+
+	var secretToUpdate *apiv1.Secret
+	{
+		TTL, err := time.ParseDuration(key.CrtTTL(customObject))
+		if err != nil {
+			return false, microerror.Mask(err)
+		}
+
+		renew, err := r.shouldCertBeRenewed(currentSecret, TTL, r.expirationThreshold)
+		if IsMissingAnnotation(err) {
```
JosephSalisbury: If the annotation is missing, could we add it?

service/resource/vaultcrt/resource.go
xh3b4sd: This is the golang naming. 

- https://golang.org/pkg/time/#Parse
- https://golang.org/pkg/time/#Time.Format

JosephSalisbury: TIL. Thanks! <3

service/resource/vaultcrt/update.go
xh3b4sd: We add it with new clusters. I think there will be a transition period like in the kvm-operator with the version bundle versions tracked in the deployments.

JosephSalisbury: My thinking was we could avoid the transition period. No strong opinion.

PR #120, opened 2017-11-27
Add system:masters org to API cert

rossf7:
Adds the system:masters org to match the configuration created by kubernetesd and is needed for RBAC on locally launched clusters.

https://github.com/giantswarm/kubernetesd/blob/e74b3ac4ea48c6b24b88d1935c02dde79ac07ca9/service/creator/certificate.go#L68

### DIFF
--- a/examples/cert-resource-lab-chart/templates/api-cert.yaml
+++ b/examples/cert-resource-lab-chart/templates/api-cert.yaml
@@ -21,3 +21,5 @@ spec:
   - "172.31.0.1"
   ttl: "720h"
   allowBareDomains: true
+  organizations:
+  - "system:masters"


fgimenez:
Thanks a lot for this, there's a problem with cert-resource-lab-chart though, it can't be automatically built and we can't reuse it from other projects. That's why I created https://github.com/giantswarm/cert-resource-lab, the fix you add in this PR is already there. 

I have pending a PR for removing `example/cert-resource-lab-chart` and updating the local example doc, will be ready very soon, pls bear with me :)

rossf7:
Ah I'd forgotten about the new repo. No worries, closing since this isn't needed.

PR #122, opened 2017-11-29
versioned resources

xh3b4sd:


service/keyv1/error.go
```diff
@@ -1,4 +1,4 @@
-package key
+package keyv1
```
xh3b4sd: Relevant packages got renamed from `foo` to `foov1`. Nothing else changed. Only renaming. Background is the migration to CRDs where we are going to introduce `foov2` because of the type change for the custom object.

PR #127, opened 2017-12-01
added watch for v2 resources

xh3b4sd:


service/framework.go
```diff
@@ -17,18 +22,279 @@ import (
...
+	"github.com/giantswarm/cert-operator/service/resource/vaultcrtv2"
 	"github.com/giantswarm/cert-operator/service/resource/vaultpkiv1"
+	"github.com/giantswarm/cert-operator/service/resource/vaultpkiv2"
 )
 
 const (
 	ResourceRetries uint64 = 3
 )
 
+func newCRDFramework(config Config) (*framework.Framework, error) {
```
xh3b4sd: This is the new framework to watch CRDs.

service/framework.go
```diff
@@ -17,18 +22,279 @@ import (
...
+
+		vaultRole, err = vaultrole.New(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var vaultCrtResource framework.Resource
+	{
+		c := vaultcrtv2.DefaultConfig()
```
xh3b4sd: We use the v2 resources here.

service/framework.go
```diff
@@ -17,18 +22,279 @@ import (
...
+
+		metricsWrapConfig := metricsresource.DefaultWrapConfig()
+		metricsWrapConfig.Name = config.Name
+		resources, err = metricsresource.Wrap(resources, metricsWrapConfig)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var clientSet *versioned.Clientset
```
xh3b4sd: This is our generated clientset.

service/framework.go
```diff
@@ -17,18 +22,279 @@ import (
...
+		clientSet, err = versioned.NewForConfig(c)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+
+	var newWatcherFactory informer.WatcherFactory
+	{
+		newWatcherFactory = func() (watch.Interface, error) {
+			watcher, err := clientSet.CoreV1alpha1().CertConfigs("").Watch(apismetav1.ListOptions{})
```
xh3b4sd: We use the generated client to watch custom objects of the CRD.

service/resource/vaultcrtv1/metric.go
```diff
@@ -19,7 +19,3 @@ var versionBundleVersionGauge = prometheus.NewGaugeVec(
 	},
 	[]string{"major", "minor", "patch"},
 )
-
-func init() {
-	prometheus.MustRegister(versionBundleVersionGauge)
-}
```
xh3b4sd: There are multiple resources for different versions who emit the same metric. For now I just removed the registration in v1 since this is going away anyway. I am wondering if there is a more graceful way to register metrics multiple times.

service/service.go
```diff
@@ -158,6 +168,7 @@ func New(config Config) (*Service, error) {
 
 func (s *Service) Boot() {
 	s.bootOnce.Do(func() {
-		s.CustomObjectFramework.Boot()
+		go s.CRDFramework.Boot()
+		go s.CustomObjectFramework.Boot()
```
xh3b4sd: Here we boot the different frameworks asynchronously so they do not block each other.

xh3b4sd:
Successfully tested on lycan.

service/resource/vaultcrtv1/metric.go
JosephSalisbury: I reckon we'll probably need to version the metrics :/

xh3b4sd: That would be weird because this means we would lose historical data and have to hustle a lot with keeping dashboards in sync and stuff. Maybe we could wildcard metrics in rge dashboards but I would probably prefer not to. Uncertain.

PR #131, opened 2017-12-08
Change wip to false

oponder:
Otherwise we will not be able to make a non wip release bundle.


oponder:
and kvm-operator too i think... but i'm not seeing ingress-operator in the list of release endpoints

xh3b4sd:
Right, ingress does not have this yet, but KVM.

PR #134, opened 2017-12-11
enabled TPR migration code

xh3b4sd:


MarcelMue:
The e2e test failed here, is it running fine on geckon?

xh3b4sd:
The e2e magic does not run on geckon. It runs on AWS AFAIK. The integration tests fail because I disabled the TPR reconciliation. This needs a proper rewrite. I would like to do this separately later on.

PR #135, opened 2017-12-11
execute migration asynchronously

xh3b4sd:


service/framework.go
```diff
@@ -261,7 +261,7 @@ func newCRDFramework(config Config) (*framework.Framework, error) {
 	}
 
 	// TODO remove after migration.
-	migrateTPRsToCRDs(config.Logger, clientSet)
+	go migrateTPRsToCRDs(config.Logger, clientSet)
```
xh3b4sd: It turns out we have so many custom objects that the migration takes too long for k8s to get some healthy response from the health endpoints. This should do the trick.

xh3b4sd:
Going with this to fix alerts.

PR #139, opened 2017-12-12
use apiextensions in e2e test

fgimenez:
Uses an interface that unifies functionality from the common clientset and from our api extensions.

### DIFF
--- a/integration/client.go
+++ b/integration/client.go
@@ -4,40 +4,19 @@ package integration
 
 import (
 	giantclientset "github.com/giantswarm/apiextensions/pkg/clientset/versioned"
-	corev1alpha1 "github.com/giantswarm/apiextensions/pkg/clientset/versioned/typed/core/v1alpha1"
 	"github.com/giantswarm/microerror"
 	"k8s.io/client-go/kubernetes"
-	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
-	extensionsv1beta1 "k8s.io/client-go/kubernetes/typed/extensions/v1beta1"
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/giantswarm/e2e-harness/pkg/harness"
 )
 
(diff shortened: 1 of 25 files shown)


integration/client.go
```diff
@@ -3,14 +3,41 @@
...
 	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/giantswarm/e2e-harness/pkg/harness"
 )
 
-func getK8sClient() (kubernetes.Interface, error) {
+type k8sClient interface {
+	CoreV1() corev1.CoreV1Interface
+	ExtensionsV1beta1() extensionsv1beta1.ExtensionsV1beta1Interface
+	CoreV1alpha1() corev1alpha1.CoreV1alpha1Interface
```
kopiczko: What will happen when we release `CoreV1`?

fgimenez:
Ok, I've updated it to use a client's struct, pls take another look. IMO with this approach we lose the help of the type system (we can pass nil or uninitialized members, that's why we see so many checks on our `new` logic), the signature of the methods is much less expressive this way, but of course I can be wrong, will keep trying to convince you guys anyway! :)

integration/client.go
fgimenez: I would iterate, applying the same approach again. At some point our `CoreV1` would differ from upstream's, there I would define the interface and an implementer type that decides which implementation to use for each of the requested methods.

fgimenez: This way we can use a single client for interacting with k8s, which is good for the legibility and maintainability of code (the client code depend on who implements a resource IMO, maybe that can change in the future).

kopiczko:
You can also pass nil or uninitialized members to a function call. Is there any difference?

fgimenez:
Sorry, I was ranting about something else in this case :D not passing the receiver as parameter but struct with embedded dependencies as parameters.

PR #141, opened 2018-01-11
moved to recent k8s dependencies

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/1554. Here we remove migration code and the `certificatetpr` dependency.


integration/operator_test.go
```diff
@@ -110,7 +109,7 @@ func (cs *clients) tearDown() {
 	runCmd("helm delete cert-operator --purge")
 	runCmd("helm delete cert-resource-lab --purge")
 	cs.K8sCs.CoreV1().Namespaces().Delete("giantswarm", &metav1.DeleteOptions{})
-	cs.K8sCs.ExtensionsV1beta1().ThirdPartyResources().Delete(certificatetpr.Name, &metav1.DeleteOptions{})
+	cs.K8sCs.ExtensionsV1beta1().ThirdPartyResources().Delete("certificate.giantswarm.io", &metav1.DeleteOptions{})
```
xh3b4sd: @fgimenez I want to get rid of the `certificatetpr` dependency. I just hardcoded this now. I think this deserves some love anyway because we are not going to do anything with TPRs anymore.

service/framework.go
```diff
@@ -255,28 +248,13 @@ func newCRDFramework(config Config) (*framework.Framework, error) {
...
-		}
-	}
-
 	var newInformer *informer.Informer
 	{
-		informerConfig := informer.DefaultConfig()
+		c := informer.DefaultConfig()
 
-		informerConfig.WatcherFactory = newWatcherFactory
+		c.Watcher = clientSet.CoreV1alpha1().CertConfigs("")
```
xh3b4sd: This is because of the vendor update. Cert-operator was quite behind.

service/keyv2/key.go
```diff
@@ -7,6 +7,20 @@ import (
...
+	KeyID = "key"
+)
+
+const (
+	// ComponentLabel is the label used in the secret to identify a cluster
+	// component.
+	ComponentLabel string = "clusterComponent"
+	// ClusterIDLabel is the label used in the secret to identify a cluster.
+	ClusterIDLabel string = "clusterID"
+)
```
xh3b4sd: This is copy/paste from `certificatetpr`. Good enough for me now. We can boyscout this later.

integration/operator_test.go
```diff
@@ -13,10 +13,9 @@ import (
 	"time"
 
 	"github.com/cenk/backoff"
-	"github.com/giantswarm/certificatetpr"
 	"github.com/giantswarm/microerror"
+	apiv1 "k8s.io/api/core/v1"
```
kopiczko: JFYI I saw it usually aliased to `corev1`.

service/keyv2/key.go
kopiczko: We can take it from `certificates` repo.

integration/operator_test.go
xh3b4sd: I noticed this is inconsistent across the project. So far I simply kept `apiv1` because this is what it was before. We can also boyscout that later in case there is need.

PR #142, opened 2018-01-12
moved to use dep

xh3b4sd:


xh3b4sd:
The tests fail because the integration tests fail which need love. This is because nobody took care about the CRD magic after we moved from TPRs. The unit tests are fine.

fgimenez:
AFAIK the e2e tests are using CRDs https://github.com/giantswarm/cert-operator/blob/master/integration/operator_test.go#L229, they were passing this morning https://circleci.com/workflow-run/a5d2ed7f-2303-4b92-9a8b-5ffe7e4e57e2. I can take a look at the failure, looks like there's a compilation error.

xh3b4sd:
Then I fucked it up actually, I just don't understand what is missing.

fgimenez:
master should be fine https://github.com/giantswarm/cert-operator/pull/143

PR #147, opened 2018-01-22
added basic implementation for GetDesiredState

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2362.

### DIFF
--- a/service/resource/vaultrolev1/current.go
+++ b/service/resource/vaultrolev1/current.go
@@ -3,9 +3,10 @@ package vaultrolev1
 import (
 	"context"
 
-	"github.com/giantswarm/cert-operator/service/keyv2"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/vaultrole"
+
+	"github.com/giantswarm/cert-operator/service/keyv2"
 )
 
 func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interface{}, error) {
--- a/service/resource/vaultrolev1/desired.go
+++ b/service/resource/vaultrolev1/desired.go
@@ -2,8 +2,37 @@ package vaultrolev1
 
 import (
 	"context"
+	"time"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/vaultrole"
+
+	"github.com/giantswarm/cert-operator/service/keyv2"
 )
 
 func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interface{}, error) {
-	return nil, nil
+	customObject, err := keyv2.ToCustomObject(obj)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


service/resource/vaultrolev1/current.go
```diff
@@ -3,9 +3,10 @@ package vaultrolev1
 import (
 	"context"
 
-	"github.com/giantswarm/cert-operator/service/keyv2"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/vaultrole"
+
+	"github.com/giantswarm/cert-operator/service/keyv2"
```
xh3b4sd: Boyscouting.

service/resource/vaultrolev1/desired.go
```diff
@@ -2,8 +2,37 @@ package vaultrolev1
...
+	}
+
+	role := &vaultrole.Role{
+		AllowBareDomains: keyv2.AllowBareDomains(customObject),
+		AllowSubdomains:  AllowSubdomains,
+		AltNames:         keyv2.AltNames(customObject),
+		ID:               keyv2.ClusterID(customObject),
+		Organizations:    keyv2.Organizations(customObject),
+		TTL:              TTL,
+	}
```
xh3b4sd: That is basically it.

service/resource/vaultrolev1/desired_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"testing"
+	"time"
+
+	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/micrologger/microloggertest"
+	"github.com/giantswarm/vaultrole"
+	"github.com/giantswarm/vaultrole/vaultroletest"
+)
+
+func Test_Resource_VaultRole_GetDesiredState(t *testing.T) {
```
xh3b4sd: And some fancy tests.

service/resource/vaultrolev1/resource.go
```diff
@@ -8,7 +8,10 @@ import (
 )
 
 const (
-	Name = "vaultrolev1"
+	// AllowSubdomains defines whether to allow the generated root CA of the PKI
+	// backend to allow sub domains as common names.
+	AllowSubdomains = true
```
xh3b4sd: This seems to be the standard here. We do not configure it via CROs.

kopiczko: I don't see a reason to extract a constant. I'd just describe variable inside a function to not jump around.

Also I would make those constants private.

xh3b4sd: I always like this for godoc reasons. When you check the package you directly see that it is the hardcoded default.

PR #151, opened 2018-01-23
use config hash to detect updates

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2362.

### DIFF
--- a/service/keyv2/key.go
+++ b/service/keyv2/key.go
@@ -1,6 +1,8 @@
 package keyv2
 
 import (
+	"crypto/sha1"
+	"encoding/json"
 	"fmt"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
@@ -43,6 +45,19 @@ func CommonName(customObject v1alpha1.CertConfig) string {
 
 func CrtTTL(customObject v1alpha1.CertConfig) string {
 	return customObject.Spec.Cert.TTL
+}
+
+func CustomObjectHash(customObject v1alpha1.CertConfig) (string, error) {
+	b, err := json.Marshal(customObject.Spec.Cert)
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	h := sha1.New()
+	h.Write(b)
+	bs := h.Sum(nil)
+
+	return fmt.Sprintf("%x", bs), nil
 }
 
 func IPSANs(customObject v1alpha1.CertConfig) []string {
--- a/service/resource/vaultcrtv2/desired.go
+++ b/service/resource/vaultcrtv2/desired.go
@@ -19,6 +19,11 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 
 	r.logger.Log("cluster", keyv2.ClusterID(customObject), "debug", "computing the desired secret")
 
+	hash, err := keyv2.CustomObjectHash(customObject)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
 	// NOTE that the actual secret content here is left blank because only the
 	// issuer backend, e.g. Vault, can generate certificates. This has to be
 	// considered when computing the create, delete and update state.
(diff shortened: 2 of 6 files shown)


service/keyv2/key.go
```diff
@@ -45,6 +47,19 @@ func CrtTTL(customObject v1alpha1.CertConfig) string {
 	return customObject.Spec.Cert.TTL
 }
 
+func CustomObjectHash(customObject v1alpha1.CertConfig) (string, error) {
+	b, err := json.Marshal(customObject.Spec.Cert)
```
xh3b4sd: So the idea is to use the config hash to detect changes.

service/resource/vaultcrtv2/desired.go
```diff
@@ -19,13 +19,19 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
...
+	}
+
 	// NOTE that the actual secret content here is left blank because only the
 	// issuer backend, e.g. Vault, can generate certificates. This has to be
 	// considered when computing the create, delete and update state.
 	secret := &apiv1.Secret{
 		ObjectMeta: apismetav1.ObjectMeta{
 			Name: keyv2.SecretName(customObject),
 			Annotations: map[string]string{
+				ConfigHashAnnotation:           hash,
```
xh3b4sd: We set the hash as annotation to the secret.

service/resource/vaultcrtv2/update.go
```diff
@@ -103,25 +103,52 @@ func (r *Resource) newUpdateChange(ctx context.Context, obj, currentState, desir
...
+			return false, microerror.Maskf(missingAnnotationError, "current secret")
+		}
+		d, ok := desiredSecret.Annotations[ConfigHashAnnotation]
+		if !ok {
+			return false, microerror.Maskf(missingAnnotationError, "desired secret")
+		}
+
+		if c != d {
+			return true, nil
+		}
```
xh3b4sd: When the current hash and the desired hash differ, the update is done. This causes cert config changes to renew certificates automatically. Note that the Vault role is already updated at this point in case it changed due to cert config changes. This is already handled in the role resource.

service/resource/vaultcrtv2/update_test.go
```diff
@@ -270,7 +283,128 @@ func Test_Resource_VaultCrt_shouldCertBeRenewed(t *testing.T) {
 			}
 		}
 
-		result, err := newResource.shouldCertBeRenewed(tc.Secret, tc.TTL, tc.Threshold)
+		result, err := newResource.shouldCertBeRenewed(tc.Secret, tc.Secret, tc.TTL, tc.Threshold)
```
xh3b4sd: The desired secret does not matter here so we simply reuse the given one because we need a kind of useful one.

service/resource/vaultcrtv2/update_test.go
```diff
@@ -270,7 +283,128 @@ func Test_Resource_VaultCrt_shouldCertBeRenewed(t *testing.T) {
...
+					},
+				},
+			},
+			ErrorMatcher:   nil,
+			ExpectedResult: false,
+		},
+
+		// Test 3 ensures having no config hash value for the current state and
+		// having a config hash value for the desired state results in updating the
+		// secret.
```
xh3b4sd: This should actually be the case we see a lot when this PR gets merged.

service/keyv2/key.go
kopiczko: I wouldn't say this is deterministic. Do you have any guarantees on field order inside json?
If there is any map inside I'm pretty sure the output is randomized on purpose.

xh3b4sd: There are only booleans strings and string slices: https://github.com/giantswarm/apiextensions/blob/f0b8bcd6122deaf8be38b9705ac528cf4b749e3a/pkg/apis/core/v1alpha1/cert_types.go#L61-L70.

kopiczko: Are you sure json fiels will be always marshaled in the same order? What happens when they aren't?

xh3b4sd: The JSON implementation is deterministic with the current types. It is always alphabetical. In case this breaks, certs are renewed more often than they should. We should improve on metrics and see when we renew too often. That would be cool. There is nothing that can horribly break out of a sudden though IMO. 

Please feel free to suggest a better approach here if necessary. I understand your concerns. I was thinking about several ways to do it and this was the best I was able to come up with in the scope of the postmortem issue.

kopiczko: > The JSON implementation is deterministic with the current types. It is always alphabetical.

When that is true this is fine. I just wanted to make sure you are aware of that.

Also if renewing certs is not a problem, then I'm 100% happy.

kopiczko: Thanks for the discussion.

PR #168, opened 2018-03-22
Update vaultrole dependency

tuommaki:
New Vault release (at least for us 0.9.3) brought changes in its API for
certificate AltName formatting. New vaultrole dependency has fixes for
this.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -200,7 +200,7 @@
     "key",
     "vaultroletest"
   ]
-  revision = "277001d5f551cfad1dfa15b19ddaea97e24e5cda"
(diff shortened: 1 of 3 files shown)


tuommaki:
Tested on `gauss` and `geckon` by deploying this version, launching a cluster, running an alpine container and fetching stuff from internet. All good.

tuommaki:
Just found problem with latest vaultrole on `geckon`. Somehow didn't catch this earlier, but combination of new cluster-service, flannel-operator and cert-operator produced state where `organization` actually also contains `[]interface{}` -> New PR for vaultrole.

tuommaki:
Ok. Now just tested this on geckon together with new cluster-service & flannel-operator and everything works.

PR #173, opened 2018-04-19
cancel vaultcrt resource on deletion

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/2993.


service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -38,10 +39,25 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+	if key.IsInDeletionState(customObject) {
+		n := key.ClusterNamespace(customObject)
+		list, err := r.k8sClient.CoreV1().Pods(n).List(metav1.ListOptions{})
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		if len(list.Items) != 0 {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "cannot finish deletion of the secret due to existing pods")
+			resourcecanceledcontext.SetCanceled(ctx)
+			r.logger.LogCtx(ctx, "level", "debug", "message", "canceling resource for custom object")
```
xh3b4sd: So in case a cluster deletion happens, we delete all secrets holding certificates. We still need the certs for draining nodes on KVM though. So as long as pods are there we delay the deletion of the secrets here in order to still use them in the kvm-operator. The impact of this for AWS and Azure is zero. When listing on namespaces that do not exist we get an empty list and thus do nothing here. For KVM, as soon as the draining was done and the pods got removed we get an empty list here after the delete event got replayed and remove the secrets.

service/controller/v2/resources/vaultcrt/delete.go
```diff
@@ -10,18 +10,6 @@ import (
...
-	if err != nil {
-		return nil, microerror.Mask(err)
-	}
-
-	patch := controller.NewPatch()
-	patch.SetDeleteChange(delete)
-
-	return patch, nil
-}
-
```
xh3b4sd: Boyscouting, alphabetical order.

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -7,10 +7,11 @@ import (
 
 	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
 	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
 	"github.com/prometheus/client_golang/prometheus"
-	apiv1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	apismetav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
xh3b4sd: Boyscouting, just getting the imports a bit more straight.

service/controller/v2/resources/vaultcrt/current.go
MarcelMue: I get the reasoning here, I would just really appreciate some comments in here. I doubt that I would remember this if I was reading this two weeks from now.

xh3b4sd: Done.

PR #176, opened 2018-04-27
Vendor update latest operatorkit

MarcelMue:
Towards https://github.com/giantswarm/giantswarm/issues/3078


service/controller/cert.go
```diff
@@ -133,6 +133,7 @@ func NewCert(config CertConfig) (*Cert, error) {
 	var newInformer *informer.Informer
 	{
 		c := informer.Config{
+			Logger:  config.Logger,
```
MarcelMue: This is the new dependency added, so we can have some logs while collecting metrics.

PR #186, opened 2018-06-13
Use release time to force pod redeploys

oponder:
Towards: https://github.com/giantswarm/giantswarm/issues/3395

### DIFF
--- a/helm/cert-operator-chart/templates/deployment.yaml
+++ b/helm/cert-operator-chart/templates/deployment.yaml
@@ -15,7 +15,7 @@ spec:
       labels:
         app: cert-operator
       annotations:
-        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


helm/cert-operator-chart/templates/deployment.yaml
```diff
@@ -16,6 +16,8 @@ spec:
         app: cert-operator
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
```
tuommaki: Remove checksum?

helm/cert-operator-chart/templates/deployment.yaml
```diff
@@ -16,6 +16,8 @@ spec:
...
+      annotations:
```
tuommaki: Also, there are two `annotations` now.

oponder: thx for checking closely! saved my butt :)

PR #188, opened 2018-07-09
Update vault dependency to 0.10.3

indyfree:
Towards https://github.com/giantswarm/giantswarm/issues/3621.

See also https://github.com/giantswarm/vaultrole/pull/14, https://github.com/giantswarm/vaultcrt/pull/6, https://github.com/giantswarm/vaultpki/pull/4, https://github.com/giantswarm/vaulthealthz/pull/8.

Update vault to `0.10.3` to align with our infrastructure.

The underlying libs are currently fixed to use the same vault version.


indyfree:
Tested on ginger

xh3b4sd:
What did you test though? Is there something we should write more e2e tests for? Also, in which state is the e2e test setup? Do we have to update Vault there as well?

indyfree:
I tested if secret/cert creation went well on ginger after deploy. This is also what the e2e does.  E2e vault doesn't have to be seperatly updated AFAIK. Also checked back with @rossf7 if theres more to it, but should be all fine. Do you have anything in mind?

indyfree:
Q: Do I need to update the changelog? Or only when new e.g. `v3` release?
https://github.com/giantswarm/cert-operator/blob/3cfb6d056ebc74892dcec9256105d97eb904b555/service/controller/v2/version_bundle.go#L16-L20

/cc @fgimenez partner in crime

fgimenez:
mmh AIUI these changes should be part of a new version bundle version, but according to the release index (and the code in this repo), looks like cert-operator is stuck at `0.1.0` and there are no plans to introduce new bundles (not sure about this). 

If that's the case, IMO it is ok to change the existing `v2` (`0.1.0`) and update the Vault version in the component list, it would reflect what we have in this concrete version bundle version.

indyfree:
Tested this multiple times. Going with this.

PR #194, opened 2018-08-24
removes vault health service

xh3b4sd:
Fixes https://github.com/giantswarm/giantswarm/issues/3547.


Gopkg.toml
```diff
@@ -22,12 +22,12 @@
 
 
 [[constraint]]
-  name = "github.com/cenkalti/backoff"
```
xh3b4sd: Vendoring our new backoff.

main.go
```diff
@@ -47,7 +47,7 @@ func main() {
 
 			serviceConfig.Description = description
 			serviceConfig.GitCommit = gitCommit
-			serviceConfig.Name = name
+			serviceConfig.ProjectName = name
```
xh3b4sd: Boyscouting.

server/endpoint/endpoint.go
```diff
@@ -13,21 +13,15 @@ import (
...
-	return Config{
-		// Dependencies.
-		Logger:     nil,
-		Middleware: nil,
-		Service:    nil,
-	}
+// Endpoint is the endpoint collection.
+type Endpoint struct {
+	Healthz *healthz.Endpoint
+	Version *version.Endpoint
```
xh3b4sd: Boyscouting.

server/endpoint/endpoint.go
```diff
@@ -36,24 +30,27 @@ func New(config Config) (*Endpoint, error) {
...
 	{
-		healthzConfig := healthz.DefaultConfig()
-		healthzConfig.Logger = config.Logger
-		healthzConfig.Services = []healthzservice.Service{
-			config.Service.Healthz.K8s,
-			config.Service.Healthz.Vault,
+		c := healthz.Config{
+			Logger: config.Logger,
+			Services: []healthzservice.Service{
+				config.Service.Healthz.K8s,
```
tuommaki: This is still checking against k8s API. Weren't we supposed to drop those as well?

xh3b4sd: Uncertain. The postmortem and the issue were only about Vault dependency. We did not have huge problems with the k8s healthcheck so far. We should maybe think about the health checking practises in general and derive some guidelines we can apply so we get it straight once and for all.

tuommaki: Roger that.

PR #199, opened 2018-10-02
extend-renew-period

calvix:
Increased renewThreshold to 90 days

Renew threshold of 14 days is in our case often too late as there might not happen update in each 14days.

 This will significantly reduce the amount of cluster which we will need `manually` renew because of expiring certs. 

towards https://github.com/giantswarm/giantswarm/issues/3913#issuecomment-418444839


helm/cert-operator-chart/templates/configmap.yaml
```diff
@@ -21,7 +21,7 @@ data:
           keyFile: ''
       resource:
         vaultCrt:
-          expirationThreshold: '336h'
+          expirationThreshold: '2160h'
```
xh3b4sd: How long is the TTL itself? It must be bigger than this threshold, otherwise the operator shits the bed. I am not sure if we have on startup validation for this, which we should think about adding.

xh3b4sd: I would say to make it more straight then and get this setting in here next to each other so we have it easier in the future to sort these things out.

calvix: So I guess this change won't happen now and we will have to manually renew clusters. Someone needs to take this properly and change the whole structure of how this work and remove it from the CRD.
I don't have time now to dig deep into this.

xh3b4sd: I am not sure we talk about the same thing. Lets maybe hang if you like. I was not suggesting to block this or do heavy refactoring here. I just thought it makes sense to set the configuration here like we do with the expiration threshold so we have these two numbers here in the configmap. Further I suggested to have some validation of the configuration at runtime. All this is not too big.

calvix: Currently, the expiration is set on CRD, so that would mean adjust CRD structure,  make sure migration of CRD is done, remove the setting of TTL in cluster operator and cluster service, check that all the other components don't use TTL anywhere in the code, change how cert-operator uses TTL.
I see at least 5-10 PRs, so in my humble opinion, this is a big change.

xh3b4sd: I see what you mean. I thought there is some default we use and that is configured in the operator anyway. All good.

calvix: @xh3b4sd Do you want me to create an issue for that in sig-operator?

xh3b4sd: No, I thought it works in a different way. All good. Thank you.

PR #200, opened 2018-10-02
Add CRD label selector for e2e tests

tuommaki:
On KVM e2e tests there are multiple instances of operators running at
the same time. In order to prevent race conditions, add label selector
to informer to only process CRs meant for current instance of operator.

Towards https://github.com/giantswarm/giantswarm/issues/4196#issuecomment-421297710

### DIFF
--- a/helm/cert-operator-chart/values.yaml
+++ b/helm/cert-operator-chart/values.yaml
@@ -1,1 +1,2 @@
+labelSelector: ''
 namespace: giantswarm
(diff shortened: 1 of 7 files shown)


helm/cert-operator-chart/values.yaml
```diff
@@ -1 +1,2 @@
+labelSelector: ''
```
xh3b4sd: Can we have this? I would like to get the values aligned with the actual config maps and flags. 

```
service: 
  crd: 
    labelSelector: ''
```

tuommaki:
I changed the structure as suggested (and approved for `flannel-operator`). PTAL :)

tuommaki:
Tested on `gorgoth`

PR #202, opened 2018-10-09
 adds separate collector service to measure Vault token expiration

xh3b4sd:
We had https://github.com/giantswarm/giantswarm/issues/4291. I noticed this also affects `cert-operator`. So this is basically copy/paste.


service/collector/vault.go
```diff
@@ -0,0 +1,82 @@
...
+	ExpireTimeKey = "expire_time"
+	// ExpireTimeLayout is the layout used for time parsing when inspecting the
+	// expiration date of the used Vault token. This layout is specific to Vault
+	// as they define it.
+	ExpireTimeLayout = "2006-01-02T15:04:05"
+)
+
+var (
+	tokenExpireTimeDesc = prometheus.NewDesc(
+		prometheus.BuildFQName("cert_operator", "vault", "token_expire_time"),
```
xh3b4sd: The whole collector is basically the same as in `cluster-service`, but this metric name difference.

xh3b4sd: Note that I need to add an alerting rule in `g8s-prometheus` as well still.

PR #204, opened 2018-10-11
Make clusterrole names configurable

MarcelMue:
Same old as in kvm-operator and flannel-operator.

towards giantswarm/giantswarm#4388


integration/operator_test.go
```diff
@@ -32,6 +32,10 @@ Installation:
         Address: http://vault.default.svc.cluster.local:8200
         CA:
           TTL: 1440h
+    GiantSwarm:
+      CertOperator:
+        CRD:
+          LabelSelector: 'giantswarm.io/cluster={{ .ClusterName }}'
```
MarcelMue: I added this in an attempt to make the e2e test green, but I honestly have no idea why it is failing. The output tells me that something exists and right afterwards tells me that it can't be found. The repo used to create those resources doesn't exist anymore. I created an issue about my feelings here:
https://github.com/giantswarm/giantswarm/issues/4412

MarcelMue: I do not want to deal with this right now. Any ideas how to proceed?

PR #216, opened 2019-07-11
Follow the metric name suggestion

tomahawk28:
From following this suggestion, I also found out cert-operator needs to change the metric name in here. 

# TODO: I'll make another PR in g8s-Prometheus so we can mitigate zero time.


service/controller/v2/resources/cleansing/resource.go
```diff
@@ -0,0 +1,16 @@
+package cleansing
```
MarcelMue: What is this about? I get the rename but I don't follow this.

tomahawk28: sorry, my bad, somehow it goes with initial commit.

PR #219, opened 2019-07-30
Skip checking running pods when deleting non-supporting certconfigs.

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/6442

We will skip checking running pods on cluster namespace if only that `CertConfig` is not supporting any more.

In this particular PR, we are aiming for deletion of `calico` cert.

### DIFF
--- a/service/controller/v2/resources/vaultcrt/current.go
+++ b/service/controller/v2/resources/vaultcrt/current.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/certs"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/operatorkit/controller/context/finalizerskeptcontext"
 	"github.com/giantswarm/operatorkit/controller/context/resourcecanceledcontext"
@@ -49,6 +50,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
+
+		if !r.checkCertType(customObject) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "do not supporting this cert type any longer.")
+			return nil, nil
+		}
+
 		n := key.ClusterNamespace(customObject)
 		list, err := r.k8sClient.CoreV1().Pods(n).List(metav1.ListOptions{})
 		if err != nil {
@@ -65,6 +72,16 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	}
 
 	return secret, nil
+}
+
+func (r *Resource) checkCertType(customObject v1alpha1.CertConfig) bool {
+	c := certs.Cert(customObject.Spec.Cert.ClusterComponent)
+	for _, cert := range certs.AllCerts {
+		if cert == c {
+			return true
+		}
+	}
+	return false
 }
 
 func (r *Resource) updateVersionBundleVersionGauge(ctx context.Context, customObject v1alpha1.CertConfig, gauge *prometheus.GaugeVec, secret *corev1.Secret) {
--- a/service/controller/v2/resources/vaultcrt/current.go
+++ b/service/controller/v2/resources/vaultcrt/current.go
@@ -50,7 +50,6 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
-
 		if !r.checkCertType(customObject) {
 			r.logger.LogCtx(ctx, "level", "debug", "message", "do not supporting this cert type any longer.")
 			return secret, nil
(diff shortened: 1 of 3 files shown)


service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -49,6 +50,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
+
+		if !r.checkCertType(customObject) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "do not supporting this cert type any longer.")
```
kopiczko: Is "this cert type" a useful information?

```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", "do not supporting this cert type any longer")
```

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -49,6 +50,12 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
+
```
kopiczko:
```suggestion
```

service/controller/v2/resources/vaultcrt/current.go
tomahawk28: If we see the message through luigi it will display as below, 
```
default/certconfigs/h0cg9-calico vaultcrtv2.GetCurrentState do not supporting this cert type any longer
```
I think it makes sense to arrange it with `this cert type` so you could guess it as `calico` type, WDYT?

kopiczko: Ah yes because this reconciles the cert. That's fine then.

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -49,6 +50,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
+		if !r.checkCertType(customObject) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "do not supporting this cert type any longer.")
```
kopiczko:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", "unsupported cert type")
```

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -49,6 +50,11 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
...
+			return secret, nil
```
kopiczko: Should the whole reconciliation or at least resource be cancelled?

tomahawk28: Since we need to keep going ahead with the deletion process, I don't think we could cancel any kind of things in here, neither reconciliation nor resource.

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -67,6 +73,16 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	return secret, nil
 }
 
+func (r *Resource) checkCertType(customObject v1alpha1.CertConfig) bool {
```
rossf7: I'm having a hard time following this. So `certs.AllCerts` are the certs we support now which excludes Calico?

I'd like to see some godoc here on what we are checking and why we are checking it.

tomahawk28: @rossf7 I added godoc on private function and just above the function usage too, hope that helps.

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -67,6 +75,17 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	return secret, nil
 }
 
+// checkCertType checks whether customObject is one of the Cert types we are supporting in certs library.
+func (r *Resource) checkCertType(customObject v1alpha1.CertConfig) bool {
+	c := certs.Cert(customObject.Spec.Cert.ClusterComponent)
```
rossf7:
```suggestion
	c := certs.Cert(key.ClusterComponent(customObject))
```

I think we should have a key function for this and also use it in the logging.

service/controller/v2/resources/vaultcrt/current.go
```diff
@@ -49,6 +50,13 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 	// draining was done and the pods got removed we get an empty list here after
 	// the delete event got replayed. Then we just remove the secrets as usual.
 	if key.IsDeleted(customObject) {
+		// If this customObject is not the cert we are supporting in certs library,
+		// we don't need to check for running pods.
+		if !r.checkCertType(customObject) {
+			r.logger.LogCtx(ctx, "level", "debug", "message", "unsupported cert type")
```
rossf7:
```suggestion
			r.logger.LogCtx(ctx, "level", "debug", "message", "unsupported cert type %#q", key.ClusterComponent(customObject))
```

I think we should log the component so it's clear what is going on.

tomahawk28: @rossf7 Thanks, I added unittest as well.

tomahawk28:
Saw this log on `centaur` 
```D 07/30 15:16:11 vaultcrtv2 unsupported cert type %#q | cert-operator/service/controller/v2/resources/vaultcrt/current.go:56 | /apis/core.giantswarm.io/v1alpha1/namespaces/default/certconfigs/h0cg9-calico=version | 183893320=(MISSING) | 8=event | GetCurrentState=controller | calico=resource | cert-operator=loop | delete=object | vaultcrtv2=function```
🤦‍♂ I'll come up with another PR.

PR #221, opened 2019-08-12
handle vault updates more graceful internally

xh3b4sd:
When we update Vault on the control planes the operator throws a couple of errors which we should simply handle a bit more graceful. 

> {"caller":"github.com/giantswarm/cert-operator/vendor/github.com/giantswarm/operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:71","controller":"cert-operator","event":"update","function":"GetCurrentState","level":"warning","loop":"91395","message":"retrying due to error","object":"/apis/core.giantswarm.io/v1alpha1/namespaces/default/certconfigs/sh0z0-app-operator-api","resource":"vaultpkiv2","stack":"[{/go/src/github.com/giantswarm/cert-operator/vendor/github.com/giantswarm/operatorkit/resource/wrapper/retryresource/crud_resource_ops_wrapper.go:64: } {/go/src/github.com/giantswarm/cert-operator/service/controller/v2/resources/vaultpki/current.go:27: } {/go/src/github.com/giantswarm/cert-operator/vendor/github.com/giantswarm/vaultpki/current.go:37: } {Get https://vault.g8s.amag.ch:8200/v1/sys/mounts: http: server gave HTTP response to HTTPS client}]","time":"2019-08-12T11:27:05.792401+00:00","underlyingResource":"vaultpkiv2","version":"172692767"}

### DIFF
--- a/service/collector/vault.go
+++ b/service/collector/vault.go
@@ -62,6 +62,7 @@ func (v *Vault) Collect(ch chan<- prometheus.Metric) error {
 	secret, err := v.vaultClient.Auth().Token().LookupSelf()
 	if IsVaultAccess(err) {
 		v.logger.LogCtx(ctx, "level", "debug", "message", "vault not reachable")
+		v.logger.LogCtx(ctx, "level", "debug", "message", "vault upgrade in progress")
 		v.logger.LogCtx(ctx, "level", "debug", "message", "canceling collection")
 		return nil
 
--- a/service/collector/vault.go
+++ b/service/collector/vault.go
@@ -1,6 +1,7 @@
 package collector
 
 import (
+	"context"
 	"strings"
 	"time"
 
@@ -56,8 +57,15 @@ func NewVault(config VaultConfig) (*Vault, error) {
 }
 
 func (v *Vault) Collect(ch chan<- prometheus.Metric) error {
+	ctx := context.Background()
+
 	secret, err := v.vaultClient.Auth().Token().LookupSelf()
-	if err != nil {
+	if IsVaultAccess(err) {
+		v.logger.LogCtx(ctx, "level", "debug", "message", "vault not reachable")
+		v.logger.LogCtx(ctx, "level", "debug", "message", "canceling collection")
+		return nil
+
+	} else if err != nil {
 		return microerror.Mask(err)
 	}
 
--- a/service/controller/cert.go
+++ b/service/controller/cert.go
@@ -52,34 +52,6 @@ func NewCert(config CertConfig) (*Cert, error) {
 	if config.G8sClient == nil {
 		return nil, microerror.Maskf(invalidConfigError, "config.G8sClient must not be empty")
 	}
-	if config.K8sClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.K8sClient must not be empty")
-	}
-	if config.K8sExtClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.K8sExtClient must not be empty")
-	}
-	if config.Logger == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.Logger must not be empty")
-	}
-	if config.VaultClient == nil {
-		return nil, microerror.Maskf(invalidConfigError, "config.VaultClient must not be empty")
-	}
-
-	if config.CATTL == "" {
-		return nil, microerror.Maskf(invalidConfigError, "config.CATTL must not be empty")
-	}
-	if config.CommonNameFormat == "" {
-		return nil, microerror.Maskf(invalidConfigError, "config.CommonNameFormat must not be empty")
-	}
-	if config.ExpirationThreshold == 0 {
-		return nil, microerror.Maskf(invalidConfigError, "config.ExpirationThreshold must not be empty")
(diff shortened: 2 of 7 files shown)


service/collector/error.go
```diff
@@ -1,37 +1,23 @@
 package collector
 
 import (
+	"strings"
+
 	"github.com/giantswarm/microerror"
-	"github.com/prometheus/client_golang/prometheus"
 )
 
-var alreadyRegisteredError = &microerror.Error{
```
xh3b4sd: Dead code.

service/collector/error.go
```diff
@@ -1,37 +1,23 @@
...
-}
-
+// executionFailedError is an error type for situations where Resource execution
+// cannot continue and must always fall back to operatorkit.
+//
+// This error should never be matched against and therefore there is no matcher
+// implement. For further information see:
+//
+//     https://github.com/giantswarm/fmt/blob/master/go/errors.md#matching-errors
+//
```
xh3b4sd: Aligning with the new standard.

service/controller/cert.go
```diff
@@ -52,34 +52,6 @@ func NewCert(config CertConfig) (*Cert, error) {
 	if config.G8sClient == nil {
 		return nil, microerror.Maskf(invalidConfigError, "config.G8sClient must not be empty")
 	}
-	if config.K8sClient == nil {
```
xh3b4sd: These are rather useless and not what we want.

service/controller/v2/resource_set.go
```diff
@@ -11,66 +11,56 @@ import (
...
-		return nil, microerror.Maskf(invalidConfigError, "config.Namespace must not be empty")
-	}
-	if config.ProjectName == "" {
-		return nil, microerror.Maskf(invalidConfigError, "config.ProjectName must not be empty")
-	}
-
 	var err error
 
-	var vaultCrtResource controller.Resource
+	var vaultAccessResource controller.Resource
```
xh3b4sd: Initializing the new resource.

service/controller/v2/resource_set.go
```diff
@@ -103,10 +93,10 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 
 	var vaultRoleResource controller.Resource
 	{
-		c := vaultroleresource.DefaultConfig()
```
xh3b4sd: Boyscouting here.

service/controller/v2/resource_set.go
```diff
@@ -120,6 +110,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 	}
 
 	resources := []controller.Resource{
+		vaultAccessResource,
```
xh3b4sd: Wiring the new resource.

service/controller/v2/resource_set.go
```diff
@@ -146,17 +137,12 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
...
+		cr, err := key.ToCustomObject(obj)
 		if err != nil {
 			return false
 		}
 
-		if key.VersionBundleVersion(customObject) == VersionBundle().Version {
-			return true
-		}
-		// TODO remove this hack with the next version bundle version or as soon as
-		// all certconfigs obtain a real version bundle version.
```
xh3b4sd: This should be fine now. I checked a cluster that is over a year old and the cert configs have all the version set.

service/collector/vault.go
```diff
@@ -56,8 +57,15 @@ func NewVault(config VaultConfig) (*Vault, error) {
 }
 
 func (v *Vault) Collect(ch chan<- prometheus.Metric) error {
+	ctx := context.Background()
+
 	secret, err := v.vaultClient.Auth().Token().LookupSelf()
-	if err != nil {
+	if IsVaultAccess(err) {
+		v.logger.LogCtx(ctx, "level", "debug", "message", "vault not reachable")
```
kopiczko: Should we add some explanation why vault is not reachable in the log? like `vault not reachable due to upgrade in progress`?

xh3b4sd: Additional information would make sense. I am just not sure what to go for. The error and the assumption are different things. What about adding another log line `vault upgrade in progress`?

kopiczko: I prefer single log because of my personal modus operandi. I'm pretty sure I could easily miss the other log. But do what you think it's right. I approved already.

mazzy89: DOes it make sense here to collect a metric like did in other case https://github.com/giantswarm/cluster-service/commit/135da30bc9f30b6dc2831f9e669620a140d09ef4

xh3b4sd: I am going with separate logs as they are separate already. We cannot simply look at a single log line anyway. 

Good shout on the additional metrics. In this case though I think that we are fine because there is already an alert when Vault is sealed and not available. The logging is only indicative for the cancelation behaviour. I think we are fine here for now. When there will be a stronger reason to add more error metrics we can gladly do this.

xh3b4sd:
Tested on `viking`.

PR #226, opened 2019-10-31
Cleaning certconfig belongs to deleted TCs

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/7488

Deleting all remaining `certConfigs` by checking whether cluster namespace is deleted or not.

### DIFF
--- a/service/controller/v2/resources/cleaning/create.go
+++ b/service/controller/v2/resources/cleaning/create.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/giantswarm/microerror"
-	"github.com/giantswarm/operatorkit/controller/context/reconciliationcanceledcontext"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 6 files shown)


tomahawk28:
Question; should we revert this resource after merge?

service/controller/v2/resources/cleaning/create.go
```diff
@@ -0,0 +1,45 @@
...
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customObject, err := key.ToCustomObject(obj)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	clusterID := key.ClusterID(customObject)
+
+	if key.IsDeleted(customObject) {
```
xh3b4sd: `EnsureCreated` is never called on delete events. This here does not make sense.

service/controller/v2/resources/cleaning/create.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/operatorkit/controller/context/reconciliationcanceledcontext"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/giantswarm/cert-operator/service/controller/v2/key"
+)
+
+func (r *Resource) EnsureCreated(ctx context.Context, obj interface{}) error {
+	customObject, err := key.ToCustomObject(obj)
```
xh3b4sd: We want to call the custom object here `cr`.

service/controller/v2/resources/cleaning/resource.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Config represents the configuration used to create a new cleansing resource.
+type Config struct {
+	G8sClient versioned.Interface
+	K8sClient kubernetes.Interface
+	Logger    micrologger.Logger
+}
+
+type Resource struct {
```
xh3b4sd: Is this a permanent or temporary resource? What is the problem we try to solve?

tomahawk28: Some installation contains certconfigs that belongs to deleted TCs. 
In `geckon` you could see as below. 
```
k get certconfigs
...
6srrb-app-operator-api       135d
6srrb-cluster-operator-api   135d
8tjv9-app-operator-api       174d
d20if-app-operator-api       170d
fzcf0-app-operator-api       103d
fzcf0-cluster-operator-api   103d
n00mo-app-operator-api       147d
```
After we deleted all these remaning certconfigs, we can safely revert this change.

PR #229, opened 2019-11-29
Use k8sclient collection

xh3b4sd:
Getting https://github.com/giantswarm/operatorkit/pull/338 in.


service/service.go
```diff
@@ -6,15 +6,14 @@ import (
 	"context"
 	"sync"
 
-	"github.com/giantswarm/apiextensions/pkg/clientset/versioned"
+	"github.com/giantswarm/apiextensions/pkg/apis/core/v1alpha1"
```
kopiczko: I'd prefer to import this as corev1alpha1 but I'm not big on this.

tuommaki: I think we occasionally have used `corev1` to alias `k8s.io/api/core/v1` - that gets potentially confusing quite quickly - WDYT?

kopiczko: It doesn't confuse me but I get your point.

xh3b4sd:
Just tested cluster creation on `gauss`. Works out. Tomorrow we need to make a plan on how to proceed with everything we have.

sslavic:
Some of the work from this and https://github.com/giantswarm/cert-operator/pull/228 can be extracted, merged to master and shipped in isolation from k8s+helm+cluster-api dependency updates - e.g. migration from deprecated to new internal library APIs, and migration from deprecated k8s API groups.

The rest should be merged into single consistently named branch/PR with all k8s+helm+cluster-api dependency updates and related changes. Once we get such branches for the whole stack (libs, operators, exporters, ... - and that's where consistent branch naming helps) with passing CI builds, some will need additional testing on clusters and once confident enough that it all might work we can start merging/shipping those branches.

PR #232, opened 2019-12-12
Configure default image registry domain

sslavic:
Via https://github.com/giantswarm/cert-operator/pull/231 image registry domain was made configurable, but no default value was set, and this broke e2e tests deploying cert-operator, like azure-operator I was testing https://github.com/giantswarm/azure-operator/pull/603

This PR configures `quay.io` as default image registry domain.


sslavic:
Tested on gauss

PR #234, opened 2020-02-20
Template config struct into secret

corest:
Following https://github.com/giantswarm/opsctl/pull/552

As we still want to make it possible updating tokens with a single command, this change allows embedding just token instead of a whole yaml config from installation

Merging after updating installations repo

### DIFF
--- a/helm/cert-operator-chart/templates/secret.yml
+++ b/helm/cert-operator-chart/templates/secret.yml
@@ -4,5 +4,8 @@ metadata:
   name: cert-operator-secret
   namespace: {{ .Values.namespace }}
 type: Opaque
-data:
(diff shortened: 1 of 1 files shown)


helm/cert-operator-chart/templates/secret.yml
```diff
@@ -4,5 +4,9 @@ metadata:
...
   namespace: {{ .Values.namespace }}
 type: Opaque
-data:
-  secret.yaml: {{ .Values.Installation.V1.Secret.CertOperator.SecretYaml | b64enc | quote }}
+stringData:
+  secret.yaml: |-
+    service:
+      vault:
+        config:
+          token: {{ .Values.Installation.V1.Secret.CertOperator.Service.Vault.Config.Token }}
```
kopiczko: Do we also need to change some docs/ops recipes as well?

PR #236, opened 2020-03-17
update vendor deps

xh3b4sd:
Will test this in `gauss`.


xh3b4sd:
Cluster creation worked fine. I think this is ok.

PR #238, opened 2020-03-17
cleanup orphaned PKI backends

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/9202.


xh3b4sd:
This seems to work quite well. It is only that the pod is now restarting a little because the deletion blocks the operator startup and thus makes the liveness probe fail until all backends are deleted. On `gauss` there is some way to go. 😂 

```
$ vault secrets list | grep pki | wc -l
     603
```

xh3b4sd:
Tested on `gauss`, `ginger`, `geckon` and `godsmack`.

PR #243, opened 2020-03-24
Update architect orbs

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/9339

Update to new version of architect orbs. Fixed linting issues.

I'm doing this because I was wondering if I could remove e2etemplates from our e2e tests by installing the cert-operator from the catalog. That's why I'm pushing to the catalog.


.circleci/config.yml
```diff
@@ -23,7 +23,9 @@ jobs:
         - go-cache-v1-{{ checksum "/tmp/year" }}-{{ checksum "/tmp/month" }}
         - go-cache-v1-{{ checksum "/tmp/year" }}
 
-    - run: ./architect build
+    - run:
+        name: Build using architect
+        command: ./architect build
```
kopiczko: I don't think you need that anymore when you use `architect/go-test-legacy`. Ping @tfussell.

tfussell: I haven't had a chance to finish my work here https://github.com/giantswarm/architect-orb/pull/90. Without that, you can't save the binary built with `go-architect-legacy` so it's not usable for projects that build a binary. This is an oversight on my part since I was focused on testing it with `apiextensions`. I can try to finish it tomorrow otherwise we're going to have every project on go modules before anyone can use this job I made.

service/collector/vault.go
```diff
@@ -61,9 +61,9 @@ func (v *Vault) Collect(ch chan<- prometheus.Metric) error {
 
 	secret, err := v.vaultClient.Auth().Token().LookupSelf()
 	if IsVaultAccess(err) {
-		v.logger.LogCtx(ctx, "level", "debug", "message", "vault not reachable")
-		v.logger.LogCtx(ctx, "level", "debug", "message", "vault upgrade in progress")
-		v.logger.LogCtx(ctx, "level", "debug", "message", "canceling collection")
+		v.logger.LogCtx(ctx, "level", "debug", "message", "vault not reachable")       // nolint: errcheck
+		v.logger.LogCtx(ctx, "level", "debug", "message", "vault upgrade in progress") // nolint: errcheck
+		v.logger.LogCtx(ctx, "level", "debug", "message", "canceling collection")      // nolint: errcheck
```
xh3b4sd, line 66: This should not be necessary anymore with the latest `micrologger` changes. Same for all the other occurrences of that.

service/controller/v2/key/key.go
```diff
@@ -50,8 +50,11 @@ func CustomObjectHash(customObject v1alpha1.CertConfig) (string, error) {
 		return "", microerror.Mask(err)
 	}
 
-	h := sha1.New()
-	h.Write(b)
+	h := sha256.New()
```
xh3b4sd, line 53: Note that the implication of this is that all certificates will be rolled. Shouldn't hurt, just saying.

service/collector/vault.go
fiunchinho: Fixed.

service/controller/v2/resources/vaultcrt/delete.go
```diff
@@ -46,7 +46,7 @@ func (r *Resource) NewDeletePatch(ctx context.Context, obj, currentState, desire
 	return patch, nil
 }
 
-func (r *Resource) newDeleteChange(ctx context.Context, obj, currentState, desiredState interface{}) (interface{}, error) {
+func (r *Resource) newDeleteChange(ctx context.Context, currentState interface{}) (interface{}, error) {
```
xh3b4sd, line 49: I think this is not good. It is also not necessary anymore. We disabled the noparam check. Reason is to keep the interfaces consistent for automation reasons. Code generation will be easier. Fixing interfaces and maintaining our operators will be more easily automatable when we keep the interfaces aligned. We just had that with the go modules move where we used some scripts to adapt to the recent microerror advances.

fiunchinho: Fixed

kopiczko:
@fiunchinho: moving to architect orb for an operator is not as easy as we'd like to. It requires some other manual work. I'm in the process of writing docs for that. Sorry I didn't realise that during my previous review.

fiunchinho:
np! is there anything else?

fiunchinho:
Ok, I'll leave it for you then.

kopiczko:
We are redoing it.

PR #246, opened 2020-03-24
use architect orb

xh3b4sd:
## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,58 +1,20 @@
-orbs:
-  architect: giantswarm/architect@0.4.5
-
 version: 2.1
-jobs:
-  build:
-    machine: true
-    steps:
-    - checkout
 
-    - run: |
-        wget -q $(curl -sS -H "Authorization: token $RELEASE_TOKEN" https://api.github.com/repos/giantswarm/architect/releases/latest | grep browser_download_url | head -n 1 | cut -d '"' -f 4)
-        chmod +x ./architect
-        ./architect version
-
-    - run: |
-        date +"%Y" > /tmp/year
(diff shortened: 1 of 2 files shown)


xh3b4sd:
Hey guys I need help here. I am not sure how the Circle Config for operators should look like. Couldn't find anything in the `architect-orb` docs. Uncertain which project may already does something similar. This here is pretty much the first operator I move to the new orb.

rossf7:
@xh3b4sd I think `build` should be `architect/build` and that will fix the error.

.circleci/config.yml
```diff
@@ -1,58 +1,20 @@
...
   build:
     jobs:
+      - architect/go-test:
+          name: go-test
+          filters:
+            # Trigger job also on git tag.
+            tags:
+              only: /^v.*/
+
       - build
```
xh3b4sd:
```suggestion
      - architect/build
```

xh3b4sd:
Doesn't seem to help. Can I check somewhere where this is already done? Also, when we have the build step, do we need the testing step? Do we need the build step when we have the testing step? I lack a bit of insights here and would appreciate talking with someone who knows more. 🙈

rossf7:
> Can I check somewhere where this is already done?

Mr @kopiczko likely knows better but I think you're blazing a trail. I don't know of any operators that are not using the build step. 

My knowledge here is limited but I think the build step is needed. For example for push-to-docker so there is a binary. There should also be some build cache magic going on. Sorry I can't be more help.

xh3b4sd:
Discussed with Pawel. The operator will be flattened and Team Magic takes care of the Circle Config later. Discarding this here now.

PR #264, opened 2020-04-22
helm: use Release.Revision to make chart Helm 3 compatible

wtty-fool:


helm/cert-operator/templates/deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       labels:
         {{- include "labels.selector" . | nindent 8 }}
       annotations:
-        releasetime: {{ $.Release.Time }}
+        releaseRevision: {{ $.Release.Revision }}
```
kopiczko, line 21:
```suggestion
        releaseRevision: {{ .Release.Revision }}
```

wtty-fool: Hmm, not according to changes @rossf7 suggested: https://github.com/giantswarm/azure-operator/commit/f2f2e06a9479135c4c0134716e947ba83dd2d384

kopiczko: I talked with and `$` is a typo.

rossf7: Both work but according to the docs it should be without the `$`. 

https://helm.sh/docs/chart_template_guide/builtin_objects/

Sorry for the confusion. I'm going to fix it in my previous PRs including in our docs. 😬

wtty-fool: Thanks, lads. `$` removed, marking it as resolved.

PR #266, opened 2020-04-23
version: set to v1.1.0 to make unique app deployment possible

wtty-fool:
Merge after #267 ~~so that `thiccc` can be created from `v.1.0.0`~~  :octocat: 

- Version is a mandatory field now.
- Also contains an image tag fix
- Tested on _ginger_, succesfully deployed as `unique` app

Towards https://github.com/giantswarm/giantswarm/issues/9339

## Checklist


kopiczko:
We shouldn't have `thiccc` here. Did something change?

wtty-fool:
Cool cool cool. In that case what do I do about current `v1.0.0` tag?:
- leave it as is 
- move it up to current master **before** I merge this branch
- move it up to master **after** merge

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "cert-operator"
 	source      string = "https://github.com/giantswarm/cert-operator"
-	version            = "n/a"
+	version            = "1.1.0-dev"
```
kopiczko, line 8: According to the process it should be previous version + 0.0.1-dev:

```suggestion
	version            = "1.0.1-dev"
```

PR #272, opened 2020-04-24
watch CR with cert-operator.giantswarm.io/version label

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/9343

This PR makes cert-operator only process CR which have `cert-operator.giantswarm.io/version` (moving away from version bundle concept).
Also get rid of `cert_operator_vaultcrt_resource_version_bundle_version_total` metrics which will become outdated. 

## Checklist

- [ ] Update changelog in CHANGELOG.md.


service/controller/key/key.go
```diff
@@ -100,3 +101,7 @@ func ToCustomObject(v interface{}) (v1alpha1.CertConfig, error) {
 func VersionBundleVersion(customObject v1alpha1.CertConfig) string {
 	return customObject.Spec.VersionBundle.Version
 }
```
kopiczko, line 103: Can that beremoved?

TheoBrigitte: done

service/controller/resource_set.go
```diff
@@ -145,7 +145,7 @@ func NewResourceSet(config ResourceSetConfig) (*controller.ResourceSet, error) {
 			return false
 		}
 
-		if key.VersionBundleVersion(cr) == project.NewVersionBundle().Version {
+		if key.OperatorVersion(&cr) == project.Version() {
```
xh3b4sd, line 148: I think there is no need for the handles func anymore in case we configure the correct label selector to the controller. Just did it in `aws-operator` as well today. See https://github.com/giantswarm/aws-operator/pull/2361.

service/controller/resources/vaultcrt/desired.go
```diff
@@ -31,9 +31,8 @@ func (r *Resource) GetDesiredState(ctx context.Context, obj interface{}) (interf
 		ObjectMeta: apismetav1.ObjectMeta{
 			Name: key.SecretName(customObject),
 			Annotations: map[string]string{
-				ConfigHashAnnotation:           hash,
-				UpdateTimestampAnnotation:      r.currentTimeFactory().In(time.UTC).Format(UpdateTimestampLayout),
-				VersionBundleVersionAnnotation: key.VersionBundleVersion(customObject),
```
xh3b4sd, line 36: Shouldn't we keep some version label? We should make sure the operator is only working with their particular CRs it has authority over. Otherwise all kinds of operators modify all kinds of CRs.

xh3b4sd:
BTW IMO there is no need to request review from all Chapter PE. Was that on purpose? Alternatively let's consider individuals or other teams.

service/controller/resources/vaultcrt/desired.go
TheoBrigitte: Those are annotations. But we should have `giantswarm.io/managed-by` label at least.

TheoBrigitte: That's done. PTALA

service/controller/resource_set.go
TheoBrigitte: label selector is the way to go I agree, but I'd like to get us going here. And maybe tackle this in a different PR.

TheoBrigitte:
> BTW IMO there is no need to request review from all Chapter PE. Was that on purpose? Alternatively let's consider individuals or other teams.

I wanted more eyes on this and though I was important enough to ping the chapter.

service/controller/resource_set.go
xh3b4sd: With you. I maybe got confused with the PR title.

service/controller/resources/vaultcrt/resource.go
```diff
@@ -28,6 +28,8 @@ const (
 	// UpdateTimestampLayout is the time layout used to format and parse the
 	// update timestamps tracked in the annotations of the Kubernetes secrets.
 	UpdateTimestampLayout = "2006-01-02T15:04:05.000000Z"
+	// managedByLabel is used to flag resources create by this operator.
+	managedByLabel = "giantswarm.io/managed-by"
```
tuommaki, line 32: For consistency, would it make sense to move this to `label` package?

TheoBrigitte: make sense, done.

PR #274, opened 2020-04-28
release 0.1.0

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/10389

Set version to `0.1.0` which is the current version of cert-operator.

## Checklist

- [ ] Update changelog in CHANGELOG.md.

### DIFF
--- a/pkg/project/project.go
+++ b/pkg/project/project.go
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "cert-operator"
 	source      string = "https://github.com/giantswarm/cert-operator"
-	version            = "1.0.1-dev"
+	version            = "0.1.0"
 )
 
 func Description() string {


pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "cert-operator"
 	source      string = "https://github.com/giantswarm/cert-operator"
-	version            = "1.0.1-dev"
+	version            = "0.1.0"
```
wtty-fool, line 8:
```suggestion
	version            = "0.2.0"
```

TheoBrigitte: This won't work, this version is actually use in the reconciliation loop ([see](https://github.com/giantswarm/cert-operator/blob/master/service/controller/resource_set.go#L148)) to match against CRs and they all have `0.1.0`.

wtty-fool: Oh, right. So you want to align it back with the tag. Gotcha.

PR #319, opened 2020-08-11
Skip non ref version

corest:
## Checklist

My release action here failed https://github.com/giantswarm/cert-operator/actions/runs/203867180
Validation is correct for this case, also we skip setting a new version as we need to keep the same version

### DIFF
--- a/.github/workflows/zz_generated.create_release.yaml
+++ b/.github/workflows/zz_generated.create_release.yaml
@@ -59,6 +59,13 @@ jobs:
           fi
           echo "path=\"$path\""
           echo "::set-output name=path::${path}"
+      - name: Check if reference version
+        id: ref_version
+        run: |
+          version=$(echo $title | cut -d ' ' -f 2)
+          if [[ "${version}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
+            echo ::set-output name=refversion::true
+          fi
   install_semver:
     name: Install semver
     runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- Skip vaidation of reference versions like `0.1.0-1`.
```
stone-z, line 12:
```suggestion
- Skip validation of reference versions like `0.1.0-1`.
```

corest:
I would rather skip this workflow completely for replacement releases as it starts to be overcomplicated for this rare case. But can we skip the whole action based on a condition?

.github/workflows/zz_generated.create_release.yaml
```diff
@@ -118,8 +118,15 @@ jobs:
           echo "::add-path::/opt/bin"
       - name: Checkout code
         uses: actions/checkout@v2
+      - name: Ensure non-reference version
+        id: non_ref-version
```
kopiczko, line 122:
```suggestion
        id: non_ref_version
```

.github/workflows/zz_generated.create_release.yaml
```diff
@@ -118,8 +118,15 @@ jobs:
...
+        run: |
+          if [[ "${{ needs.gather_facts.outputs.version }}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
```
kopiczko, line 124: How about adding a new output like `update_project_go: true/false` in `gather_facts` and skipping the whole job (`update_project_go`) if it's set to false?

kopiczko:
> I would rather skip this workflow completely for replacement releases as it starts to be overcomplicated for this rare case. But can we skip the whole action based on a condition?

Tag and changelog should be updated.

.github/workflows/zz_generated.create_release.yaml
corest: It is used also in the validation before tagging so I keep it as `refversion`

.github/workflows/zz_generated.create_release.yaml
```diff
@@ -59,6 +59,13 @@ jobs:
           fi
           echo "path=\"$path\""
           echo "::set-output name=path::${path}"
+      - name: Check if reference version
+        id: ref_version
+        run: |
+          version=$(echo $title | cut -d ' ' -f 2)
+          if [[ "${version}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
+            echo ::set-output name=refversion::true
+          fi
```
kopiczko, line 68:
```suggestion
          refversion=false
          if [[ "${version}" =~ ^[0-9]+.[0-9]+.[0-9]+-[0-9]+$ ]]; then
            refversion=true
          fi
          echo "refversion =\"$refversion\""
          echo "::set-output name=refversion::$refversion"
```

PR #364, opened 2021-02-11
Update dependencies for 1.18

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/14205

Updates `cert-operator` to k8s 1.18 dependencies. The operator is already versioned as part of releases, and the `cert-operator.giantswarm.io/version` label is being applied by `cluster-operator`. We want to avoid touching existing Secrets and CertConfigs in order to avoid accidental updates for release versions with change detection based on the Secret hashes.

I've tested this so far in combination with `cluster-operator` from https://github.com/giantswarm/cluster-operator/pull/1324 (which should theoretically not actually be necessary, since old `cert-operator` versions _should_ reconcile only their own bundle versions). The logs from the new operator do not include any lines regarding including cluster IDs from clusters other than the "new" one.

I created workload clusters in several versions:
```
$ gsctl list clusters
ID      ORGANIZATION   NAME               RELEASE       CREATED
52tzv   zach           zach 14.0.0        14.0.0        2021 Feb 15, 11:09 UTC
ag95m   zach           zach 9.0.10        9.0.10        2021 Feb 15, 11:07 UTC
h8gdh   zach           zach 12.1.5        12.1.5        2021 Feb 15, 11:08 UTC
i33y7   zach           zach 9.3           9.3.9         2021 Feb 15, 10:32 UTC
opy4t   zach           zach 11.6          11.6.1        2021 Feb 15, 10:32 UTC
phop6   zach           zach 13.0.0        13.0.0        2021 Feb 15, 11:08 UTC
```

I then collected the `cert.giantswarm.io/config-hash` value from every MC Secret with a `clusterComponent` label which was one of the supported cert types from `AllCerts`:
```
$ kubectl get secrets -A -o custom-columns="NAME":.metadata.name,"HASH":".metadata.annotations.cert\.giantswarm\.io/config-hash" --selector 'clusterComponent in (api, app-operator-api, aws-operator-api, calico-etcd-client, cluster-operator-api, etcd, etcd1, etcd2, etcd3, flanneld-etcd-client, internal-api, node-operator, prometheus, prometheus-etcd-client, service-account, worker)'
NAME                           HASH
52tzv-api                      <long hash>
52tzv-app-operator-api         <long hash>
...
```

I then deployed this `cert-operator` branch (and the `cluster-operator` branch) alongside the existing `cert-operator` version `0.1.0` as part of a release (so both the "old" cert-operator and this version were running simultaneously). I created a new cluster using the new release and ran CNCF conformance on that cluster.

With the cluster created and both new operators deployed on the MC, I re-ran the collection of MC Secrets and diffed the list with the original list collected before deploying the new operators. The list of hashes is identical. This suggests that the new operator will not try to reconcile old `CertConfig`s, which is our top concern to avoid unintentional node rolling.

## Checklist


service/controller/resources/vaultcrt/current.go
```diff
@@ -78,24 +79,39 @@ func (r *Resource) GetCurrentState(ctx context.Context, obj interface{}) (interf
 // checkCertType checks whether customObject is one of the Cert types we are supporting in certs library.
 func (r *Resource) checkCertType(customObject v1alpha1.CertConfig) bool {
 	c := certs.Cert(key.ClusterComponent(customObject))
-	for _, cert := range certs.AllCerts {
+	// Temporary - removed from certs, need to figure out what to do with this long term
```
stone-z, line 82: Need to decide what to do about this

service/controller/resources/vaultcrt/desired_test.go
```diff
@@ -58,29 +60,29 @@ func Test_Resource_VaultCrt_GetDesiredState(t *testing.T) {
...
 					Name: "al9qy-worker",
 					Annotations: map[string]string{
-						ConfigHashAnnotation:           "4bf7b5296ba01161f182de54b243e1400ae6660e",
-						UpdateTimestampAnnotation:      (time.Time{}).Format(UpdateTimestampLayout),
-						VersionBundleVersionAnnotation: "0.2.0",
+						ConfigHashAnnotation:      "4bf7b5296ba01161f182de54b243e1400ae6660e",
+						UpdateTimestampAnnotation: (time.Time{}).Format(UpdateTimestampLayout),
 					},
 					Labels: map[string]string{
-						"clusterID":                 "al9qy",
```
stone-z, line 80: `ClusterID` and `ClusterComponent` have been dropped from `certs` -- is there a reason to include them on future Secrets?

pkg/project/project.go
```diff
@@ -5,7 +5,7 @@ var (
 	gitSHA             = "n/a"
 	name        string = "cert-operator"
 	source      string = "https://github.com/giantswarm/cert-operator"
-	version            = "0.1.0"
+	version            = "0.3.0"
```
stone-z, line 8: Version?

stone-z:
Would also appreciate some eyes from @giantswarm/team-rocket on this one

MarcelMue:
Can you add a nice changelog message?

MarcelMue (approved):

go.mod
```diff
@@ -3,26 +3,27 @@ module github.com/giantswarm/cert-operator
...
-	sigs.k8s.io/cluster-api v0.3.11
-	sigs.k8s.io/controller-runtime v0.5.11
+	k8s.io/api v0.18.15
+	k8s.io/apimachinery v0.18.15
+	k8s.io/client-go v0.18.15
+	sigs.k8s.io/cluster-api v0.3.13
+	sigs.k8s.io/controller-runtime v0.6.3
 )
+
+replace sigs.k8s.io/cluster-api => github.com/giantswarm/cluster-api v0.3.10-gs
```
tfussell, line 29: Latest is v0.3.13-gs.

go.mod
```diff
@@ -3,26 +3,27 @@ module github.com/giantswarm/cert-operator
 go 1.13
 
 require (
-	github.com/giantswarm/apiextensions v0.4.1
-	github.com/giantswarm/certs v0.2.0
-	github.com/giantswarm/exporterkit v0.2.0
-	github.com/giantswarm/k8sclient v0.2.0
+	github.com/giantswarm/apiextensions/v3 v3.17.0
+	github.com/giantswarm/certs/v3 v3.1.1-0.20210215105438-44a9af6ad3f6
```
tfussell, line 7: Reminder to use released version here.

pkg/project/project.go
tfussell: 1.0.0 makes sense to me even though it's not that big of a change.

tfussell:
The PR looks good. I understand that you checked that this new operator won't incorrectly reconcile old cert versions, but did you also confirm that the old operator won't reconcile the new certs (matching this operator version)?

tfussell:
Might be a good chance to resolve the conversation in https://github.com/giantswarm/cert-operator/pull/328 as no one was willing to put in the time back then.

go.mod
stone-z: Thanks - updated

go.mod
stone-z: Tagged it :white_check_mark:

stone-z:
> did you also confirm that the old operator won't reconcile the new certs (matching this operator version)

Yes, with some additional considerations.

`cluster-operator` master currently applies the `cert-operator` version from the cluster's release as both the `VersionBundle` and the label. Thus, a release using new cert-operator would use a version which wouldn't be reconciled by older operators. The old `cert-operator` cancels reconciliation of new CertConfigs. However I've also opened [this PR](https://github.com/giantswarm/cluster-operator/pull/1324) to stop adding VersionBundles altogether out of an abundance of caution. 

`cluster-operator` legacy has no concept of releases, so I'm still considering how best to wire it up to use the new `cert-operator`. Currently, if no change is made to `cluster-operator` legacy, the two `cert-operator` generations will be able to just coexist until `cluster-operator` master is usable by all providers or legacy is modified to apply the label.

> Might be a good chance to resolve the conversation in #328 as no one was willing to put in the time back then.

I'm not sure it makes sense to release 0.2.0 in addition to this. WDYT?

tfussell:
> I'm not sure it makes sense to release 0.2.0 in addition to this. WDYT?

I agree. Let's just close that PR.

> cluster-operator legacy has no concept of releases, so I'm still considering how best to wire it up to use the new cert-operator. Currently, if no change is made to cluster-operator legacy, the two cert-operator generations will be able to just coexist until cluster-operator master is usable by all providers or legacy is modified to apply the label.

This sounds good to me.

go.mod
```diff
@@ -3,26 +3,27 @@ module github.com/giantswarm/cert-operator
...
+	github.com/giantswarm/apiextensions/v3 v3.17.0
+	github.com/giantswarm/certs/v3 v3.1.1
+	github.com/giantswarm/exporterkit v0.2.1
+	github.com/giantswarm/k8sclient/v5 v5.0.0
 	github.com/giantswarm/microendpoint v0.2.0
 	github.com/giantswarm/microerror v0.3.0
 	github.com/giantswarm/microkit v0.2.2
 	github.com/giantswarm/micrologger v0.5.0
-	github.com/giantswarm/operatorkit v0.2.1
+	github.com/giantswarm/operatorkit/v4 v4.0.0
```
tfussell, line 14: I recommend using v4.2.0 as there have been a few important fixes in the last two minors.

stone-z: Good idea, included that as well

tfussell (approved):

stone-z:
Tested on Azure and KVM last week. I've now bumped the version to 1.0.0 and plan to merge + release today. LMK if there are any last thoughts

MarcelMue:
The only thing (I might have missed) is the effect of upgrading a cluster to a version using this cert-operator. But we could test this later.

stone-z:
I've run upgrade tests from 9.x to releases based on 13.x using this cert-operator. Upgrade from 12/13.x to this seems to be fine

PR #386, opened 2021-04-02
Use separate CA for etcd certificates

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/16494

The goal of this PR is to create a separate PKI on `vault` to be used for etcd certificates.
This is a security best practice that is used by kubeadm.

I agree we could probably find more elegant ways of developing this feature, but implemented like this is 100% transparent for cluster operator and the provider operators.

Do you think this is a viable solution?

## Checklist

### DIFF
--- a/service/controller/resources/vaultcrt/resource.go
+++ b/service/controller/resources/vaultcrt/resource.go
@@ -1,9 +1,11 @@
 package vaultcrt
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/core/v1alpha1"
+	"github.com/giantswarm/certs/v3/pkg/certs"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/giantswarm/vaultcrt"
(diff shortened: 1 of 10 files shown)


service/controller/resources/vaultpki/delete.go
```diff
@@ -32,16 +32,20 @@ func (r *Resource) ApplyDeleteChange(ctx context.Context, obj, deleteChange inte
...
+		ids := key.PKIIdsForCluster(key.ClusterID(customObject))
 
-		err := r.vaultPKI.DeleteBackend(key.ClusterID(customObject))
-		if err != nil {
-			return microerror.Mask(err)
-		}
+		for _, id := range ids {
+			r.logger.Debugf(ctx, "deleting the Vault PKI %s in the Vault API", id)
+
+			err := r.vaultPKI.DeleteBackend(id)
```
tuommaki, line 40: Does `DeleteBackend()` return an error if non-existing backend is tried to be deleted? i.e. can this hang here if on first run only one `id` gets deleted and second one errors but then on the next reconciliation the first id doesn't exist anymore and it gets stuck?

tuommaki (approved):
LGTM in general

service/controller/resources/vaultcrt/resource.go
```diff
@@ -103,10 +105,26 @@ func (r *Resource) Name() string {
...
+		default:
+			pkiID = key.ClusterID(customObject)
+		}
+	}
+
 	c := vaultcrt.CreateConfig{
 		AltNames:      key.AltNames(customObject),
 		CommonName:    key.CommonName(customObject),
-		ID:            key.ClusterID(customObject),
+		ID:            pkiID,
```
MarcelMue, line 127: How does this behave on cluster upgrades? E.g. a cluster is upgraded from a version without this separate CA to one with the separate CA

whites11: good point. This is nothing I'm considering so far.
Current goal Is to make cluster creation to work in a CAPI-ready way and only after that I will think about upgradability

MarcelMue: Okay - we should not merge here then until this has been tested. We have really f-ed up in the past by accidentally merging incompatible changes in cert-operator.

nprokopic:
What's the state here? Can it be merged, should it be closed?

whites11:
> What's the state here? Can it be merged, should it be closed?

I didn't even remember about this PR. Do you need it?

nprokopic:
> > What's the state here? Can it be merged, should it be closed?
> 
> I didn't even remember about this PR. Do you need it?

Nope, I was just checking and cleaning up my GitHub notifications and review, and saw this one open and wondered what is the state, if we are waiting for something or just forgot about it :smile:

whites11:
We are waiting to test it properly I guess.

PR #390, opened 2021-04-08
Bump protobuf to fix CVE

MarcelMue:


stone-z:
I think we have to bump architect-orb as well for CI to pass

stone-z (approved):

PR #483, opened 2021-12-20
Renew token

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/660 and https://github.com/giantswarm/roadmap/issues/657

This PR makes cert-operator renew its vault token at every reconciliation loop to prevent it from expiring

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


calvix (approved):
make sense, LGTM if its tested

whites11:
> make sense, LGTM if its tested

I run it and it makes the API call successfully.
It's tough to test because default TTL is 6 months and so actual renewal didn't happen but at least it shouldn't make things worse.
Will merge it, thanks for the review

PR #513, opened 2022-03-31
Update capi dependencies

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/21301

## Checklist

- [X] Update changelog in CHANGELOG.md.


service/controller/cert.go
```diff
@@ -228,20 +227,6 @@ func tenantClusterExists(k8sClient k8sclient.Interface, id string) (bool, error)
 		}
 	}
 
-	// We need to check for the legacy AzureConfig CRs on Azure environments.
-	{
```
fiunchinho, line 232: We use CAPI `Cluster` CRD on `azure` so there is no need to look for `AzureConfigs`

njuettner (approved):
nice 👍🏻, with this PR we can use the latest cert-operator in AWS v17.2.0.

whites11 (approved):

PR #518, opened 2022-04-01
Bump go module major version to v2

fiunchinho:
I forgot to bump the go module version. I'll release this as `v2.0.1`

## Checklist

- [X] Update changelog in CHANGELOG.md.


.gitignore
```diff
@@ -1,4 +1,4 @@
-cert-operator
+cert-operator*
```
fiunchinho, line 1: The `Makefile` creates binaries when building that we don't want to commit

fiunchinho:
Going with a ping.

